PR #13, opened 2019-10-03
[REVIEW] Allow using package outside of conda-build

### LINKED ISSUE #11: Activate script in conda package contains hardcoded /usr/local/cuda
Issue:
The activate script in conda package, say in https://anaconda.org/conda-forge/nvcc_linux-64/10.1/download/linux-64/nvcc_linux-64-10.1-h3d80acd_0.tar.bz2, contains hardcoded `/usr/local/cuda` and ignores `CUDA_HOME` environment variable:
```
#!/bin/bash

if [[ ! -d "/usr/local/cuda" ]]
then
    echo "Cannot find: /usr/local/cuda"
    exit 1
fi
if [[ ! -f "/usr/local/cuda/lib64/stubs/libcuda.so" ]]
then
    echo "Cannot find: /usr/local/cuda/lib64/stubs/libcuda.so"
    exit 1
fi
grep "CUDA Version 10.1" /usr/local/cuda/version.txt &>/dev/null
if [[ 0 -ne 0 ]]
then
    echo "Version of installed CUDA didn't match package"
    exit 1
fi

export CUDA_HOME="/usr/local/cuda"
export CFLAGS="${CFLAGS} -I${CUDA_HOME}/include"
export CPPFLAGS="${CPPFLAGS} -I${CUDA_HOME}/include"
export CXXFLAGS="${CXXFLAGS} -I${CUDA_HOME}/include"

```
It looks like [install_nvcc.sh](https://github.com/conda-forge/nvcc-feedstock/blob/master/recipe/install_nvcc.sh) (that creates the activate script) should escape using `CUDA_HOME` when building the conda package.

kkraus14:
Checklist

Fixes #11 

We were currently using logic to find the CUDA installation using `which nvcc`, but in trying to move this logic down into the activate script it looks like something is blowing away the path (switched to cuda-memcheck so it doesn't find our new nvcc package), so cuda-memcheck isn't found.


recipe/install_nvcc.sh
```diff
@@ -2,61 +2,66 @@
...
+fi
+
+if [[ ! -d "\${CUDA_HOME}" ]]
 then
-    echo "Cannot find: ${CUDA_HOME}"
+    echo "Cannot find: \${CUDA_HOME}"
     exit 1
 fi
-if [[ ! -f "${CUDA_HOME}/lib64/stubs/libcuda.so" ]]
+if [[ ! -f "\${CUDA_HOME}/lib64/stubs/libcuda.so" ]]
```
isuruf: Same here

recipe/meta.yaml
```diff
@@ -13,7 +13,7 @@ outputs:
     script: install_nvcc.sh
     build:
       script_env:
-        - CUDA_HOME
+        - PATH
```
jakirkham: Is there a reason not to still use `CUDA_HOME` here?

kkraus14: @jakirkham All `CUDA_HOME` logic is pushed into the activate script at this point which I believe automatically inherits the environment variables, yes?

jakirkham: Ah right. Wrong question. Why is `PATH` inherited?

kkraus14: Because CUDA installation instructions say to add `$CUDA_INSTALL_DIR/bin` to `$PATH` to allow for calling things like `nvcc` or `cuda-memcheck`. Otherwise, we'd have to go on a hunt to find the `nvcc` or `cuda-memcheck` executables.

kkraus14: See here: https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#environment-setup

recipe/install_nvcc.sh
kkraus14: I believe this is because ${CUDA_HOME} isn't be resolved earlier due to the `which cuda-memcheck` issue.

recipe/install_nvcc.sh
```diff
@@ -2,61 +2,66 @@
...
 then
-    echo "Cannot find: ${CUDA_HOME}/lib64/stubs/libcuda.so"
+    echo "Cannot find: \${CUDA_HOME}/lib64/stubs/libcuda.so"
     exit 1
 fi
-grep "CUDA Version ${PKG_VERSION}" ${CUDA_HOME}/version.txt &>/dev/null
-if [[ $? -ne 0 ]]
+grep "CUDA Version \${PKG_VERSION}" \${CUDA_HOME}/version.txt &>/dev/null
```
isuruf:
```suggestion
grep "CUDA Version ${PKG_VERSION}" \${CUDA_HOME}/version.txt &>/dev/null
```

recipe/meta.yaml
```diff
@@ -32,8 +32,6 @@ outputs:
       files:
         - test.cu
       commands:
-        # Verify the symlink to the libcuda stub library exists.
-        - test -f "$(${CC} --print-sysroot)/lib/libcuda.so"
```
isuruf: Why move this down? This should have been created by activate automatically

kkraus14: I had a brainfart and started thinking I needed to wait until after the activate is called in the tests here, will revert.

recipe/install_nvcc.sh
kkraus14: Good catch. The other option here is to detect it at runtime since there's no longer anything version specific about this script / package.

kkraus14: Other than the `cudatoolkit` run_exports

recipe/install_nvcc.sh
```diff
@@ -2,61 +2,66 @@
...
 mkdir -p "${PREFIX}/etc/conda/activate.d"
-cat > "${PREFIX}/etc/conda/activate.d/${PKG_NAME}_activate.sh" <<EOF
+cat >"${PREFIX}/etc/conda/activate.d/${PKG_NAME}_activate.sh" <<EOF
 #!/bin/bash
 
-if [[ ! -d "${CUDA_HOME}" ]]
+# Default to using \$(cuda-gdb) to specify \$(CUDA_HOME).
+if [ -z \${CUDA_HOME+x} ]
+then
+    CUDA_HOME="\$(dirname \$(dirname \$(which cuda-gdb)))"
```
kkraus14: `PATH` is being blown away somewhere in the conda-build process so `cuda-gdb` isn't found and this doesn't work during testing.

recipe/install_nvcc.sh
```diff
@@ -2,61 +2,66 @@
...
+mkdir -p "\${CONDA_ENV_SYSROOT}/lib"
+ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_ENV_SYSROOT}/lib/libcuda.so"
 EOF
 
 # Unset `CUDA_HOME` in a deactivation script.
 mkdir -p "${PREFIX}/etc/conda/deactivate.d"
-cat > "${PREFIX}/etc/conda/deactivate.d/${PKG_NAME}_deactivate.sh" <<EOF
+cat >"${PREFIX}/etc/conda/deactivate.d/${PKG_NAME}_deactivate.sh" <<EOF
 #!/bin/bash
 unset CUDA_HOME
```
isuruf: @kkraus14, this needs to be removed.

kkraus14: If a user didn't have `CUDA_HOME` set before the activate script than this should be unset. I think we just need some additional logic to track if CUDA_HOME was set before the activate script to know whether we should unset it.

kkraus14:
Turns out I didn't have `shyaml` installed locally which was making it use a non cuda Docker image, 🤦‍♂

kkraus14:
@conda-forge/nvcc this should be ready for review! cc @pearu if you want to give it a shot.

recipe/install_nvcc.sh
```diff
@@ -2,61 +2,75 @@
...
-if [[ ! -f "${CUDA_HOME}/lib64/stubs/libcuda.so" ]]
+if [[ ! -f "\${CUDA_HOME}/lib64/stubs/libcuda.so" ]]
 then
-    echo "Cannot find: ${CUDA_HOME}/lib64/stubs/libcuda.so"
+    echo "Cannot find: \${CUDA_HOME}/lib64/stubs/libcuda.so"
     exit 1
 fi
-grep "CUDA Version ${PKG_VERSION}" ${CUDA_HOME}/version.txt &>/dev/null
-if [[ $? -ne 0 ]]
+grep "CUDA Version ${PKG_VERSION}" \${CUDA_HOME}/version.txt &>/dev/null
```
pearu: Instead of redirection to `/dev/null`, use
```
grep -q "CUDA Version ${PKG_VERSION}" \${CUDA_HOME}/version.txt
```

recipe/install_nvcc.sh
```diff
@@ -2,61 +2,75 @@
...
+if [[ \$? -ne 0 ]]
```
pearu: Note that conda activate scripts are pre-processed. So `$?` may not contain the return status of the `grep` command as intended. As a fix, use:
```
if grep -q "CUDA Version ${PKG_VERSION}" \${CUDA_HOME}/version.txt; then
    :
else
    echo "Version of installed CUDA in \${CUDA_HOME} didn't match package ${PKG_VERSION} (\$?)"
    exit 1
fi
```

recipe/install_nvcc.sh
pearu: But see also the comment below.

kkraus14:
@pearu updated based on your review, @jakirkham I think this is ready for another pass on review from you

recipe/meta.yaml
```diff
@@ -32,20 +30,41 @@ outputs:
       files:
         - test.cu
       commands:
+        # Get intial state of $CUDA_HOME
+        - export TEST_CUDA_HOME_INITIAL=$CUDA_HOME
```
kkraus14: In hindsight, the environment may be activated with the scripts already at this point... Is there an easy way to grab an environment variable here before the environment is set?

jakirkham: Checking the backup environment variable? Looks like it is called `CUDA_HOME_CONDA_NVCC_BACKUP` (unless I misread things). 🙂

kkraus14: I moreso want to check that roundtripping activate / deactivate doesn't have unintended side effects and wanted to get the state of things before activating.

jakirkham: I don't know of a good way to do that.

kkraus14: How could I even save the initial environment state in a test for a package like this? The test executes with the environment activated.

jakirkham: Right this is something that would need to be done in the activation script itself. We create [the activation script during the build]( https://github.com/conda-forge/nvcc-feedstock/blob/5e21ee5bf0c8f2bf5829c21521f48c3cff8bc419/recipe/install_nvcc.sh#L13-L37 ). This way we can have access to any environment variables before we modify them. Of course this doesn't eliminate other package activation scripts causing us issues here (if that's your concern). Though I think this latter concern is unlikely to be of issue here.

recipe/install_nvcc.sh
```diff
@@ -2,61 +2,98 @@
...
-# Default to using `nvcc` to specify `CUDA_HOME`.
-if [ -z ${CUDA_HOME+x} ]
-then
-    CUDA_HOME="$(dirname $(dirname $(which nvcc)))"
-fi
-
 # Set `CUDA_HOME` in an activation script.
 mkdir -p "${PREFIX}/etc/conda/activate.d"
-cat > "${PREFIX}/etc/conda/activate.d/${PKG_NAME}_activate.sh" <<EOF
+cat >"${PREFIX}/etc/conda/activate.d/${PKG_NAME}_activate.sh" <<EOF
```
jakirkham: I don't think whitespace changes like this are needed, but I could be missing something. This happens on a few subsequent lines.

recipe/install_nvcc.sh
```diff
@@ -2,61 +2,98 @@
...
 #!/bin/bash
 
-if [[ ! -d "${CUDA_HOME}" ]]
+export CUDA_HOME_CONDA_NVCC_BACKUP="\${CUDA_HOME:-}"
+export CFLAGS_CONDA_NVCC_BACKUP="\${CFLAGS:-}"
+export CPPFLAGS_CONDA_NVCC_BACKUP="\${CPPFLAGS:-}"
+export CXXFLAGS_CONDA_NVCC_BACKUP="\${CXXFLAGS:-}"
+
+# Default to using \$(cuda-gdb) to specify \$(CUDA_HOME).
```
jakirkham: Why should we use `cuda-gdb` to do this (as opposed to `nvcc` or something else)?

recipe/install_nvcc.sh
```diff
@@ -2,61 +2,98 @@
...
+if [ -z \${CUDA_HOME+x} ]
```
jakirkham: Should this be quoted?

recipe/install_nvcc.sh
```diff
@@ -2,61 +2,98 @@
...
+    if [ -n "\$CUDA_GDB_EXECUTABLE" ]; then
+        CUDA_HOME=\$(dirname \$(dirname \$CUDA_GDB_EXECUTABLE))
+    else
+        echo "Cannot determine CUDA_HOME: cuda-gdb not in PATH"
+        exit 1
+    fi
 fi
-if [[ ! -f "${CUDA_HOME}/lib64/stubs/libcuda.so" ]]
+
+if [[ ! -d "\${CUDA_HOME}" ]]
```
jakirkham: Is this still possible given the checks above?

recipe/install_nvcc.sh
```diff
@@ -2,61 +2,98 @@
...
 then
-    echo "Cannot find: ${CUDA_HOME}/lib64/stubs/libcuda.so"
+    echo "Cannot find: \${CUDA_HOME}"
```
jakirkham: Maybe just "Cannot determine CUDA_HOME" as above (assuming this block is still needed).

recipe/install_nvcc.sh
```diff
@@ -2,61 +2,98 @@
...
+    unset CFLAGS
+fi
+
+if [ -z "\${CPPFLAGS}" ]; then
+    unset CPPFLAGS
+fi
+
+if [ -z "\${CXXFLAGS}" ]; then
+    unset CXXFLAGS
+fi
```
jakirkham: I don't think we want these given we have restored the backed up values above. Though I could be missing something.

jakirkham:
Gave a review of the build script. Hopefully that helps. 🙂

recipe/meta.yaml
jakirkham: After looking over this more closely, maybe what you want can be accomplished by manually running the activation and deactivation scripts. This would allow you to configure the outside environment as you see fit and then re-run activation to see if it behaved correctly.

recipe/install_nvcc.sh
pearu: The whole point of having the `unset`-s here is to restore the original environment exactly.

(Although unlikely) when `CXXFLAGS` was not set in the parent environment before activation then without the `unset`, after deactivation `CXXFLAGS` will be defined in the parent environment (with no value), hence conda activate/deactivate is having a sideeffect.

recipe/install_nvcc.sh
```diff
@@ -2,61 +2,98 @@
...
+
+# Remove \$(libcuda.so) shared object stub from the compiler sysroot.
+CONDA_ENV_SYSROOT="\$(\${CC} --print-sysroot)"
```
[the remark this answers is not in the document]
pearu: Hmm, what if one is using clang as a C compiler? clang does not have sysroot.

recipe/meta.yaml
```diff
@@ -32,20 +30,41 @@ outputs:
...
+          else
+            echo "CUDA_HOME is set to '$CUDA_HOME'"
+          fi
         # Try building something.
         - nvcc test.cu
+        # Cleanup after ourselves
+        - unset TEST_CUDA_HOME_INITIAL
     about:
       home: https://github.com/conda-forge/nvcc-feedstock
       license: BSD 3-Clause
```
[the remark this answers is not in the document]
jakirkham: Both are valid. In either case this is unrelated to the changes in this PR. Could you please raise a separate issue or PR on this point, @pearu?

recipe/install_nvcc.sh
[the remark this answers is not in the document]
jakirkham: This is coming from the original recipe content.

I think we need to decide what the scope is here. Generally `nvcc` wants to use `gcc` on Linux. Do we want to make it use a different compiler?

While this is somewhat related to this PR, I'd be inclined to break this out as a separate issue for follow-up.

recipe/install_nvcc.sh
jakirkham: Makes sense. Thanks. Marking this as resolved.

recipe/install_nvcc.sh
```diff
@@ -2,61 +2,98 @@
...
+export CUDA_HOME="\${CUDA_HOME}"
 export CFLAGS="\${CFLAGS} -I\${CUDA_HOME}/include"
 export CPPFLAGS="\${CPPFLAGS} -I\${CUDA_HOME}/include"
 export CXXFLAGS="\${CXXFLAGS} -I\${CUDA_HOME}/include"
+
+# Add \$(libcuda.so) shared object stub to the compiler sysroot.
+# Needed for things that want to link to \$(libcuda.so).
+# Stub is used to avoid getting driver code linked into binaries.
+
+CONDA_ENV_SYSROOT="\$(\${CC} --print-sysroot)"
```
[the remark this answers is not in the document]
jakirkham: Not sure I'm following. What is the relationship? That issue seems to be about setting `CC`, which is not what is happening here.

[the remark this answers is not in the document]
pearu: Here `CC` is assumed to contain a path to a compiler (obviously) but when `nvcc` and `go` are both used as dependencies to a third package, the `go` activate will reset `CC` to nonsense (as described in https://github.com/conda-forge/go-feedstock/issues/47). If `go` activate precedes to `nvcc` activate, the `$CC --print-sysroot` will fail with a message
```
nvcc_linux-64_activate.sh: line 46: /use/the/cgo/conda/package/instead: No such file or directory
```

I realize that this is actually a `go` issue but `go` feedstock maintainers refuse to recognize this.
Hence my suggestion is to avoid using `CC` when activating `nvcc` if possible. The build script of the third-party package can workaround the `go-CC` issue (by resetting `CC`) but the workaround will not be possible if the failure happens during the activation of dependencies.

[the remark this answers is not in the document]
jakirkham: How often are people using `go` with `nvcc`? In any event, as you say, this is an issue with the `go` feedstock.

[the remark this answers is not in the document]
pearu: OmniSciDB, for instance, is using go and nvcc.
At least to me, it is a real issue.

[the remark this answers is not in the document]
jakirkham: Still this is the wrong place to handle this. It should be fixed in the `go` package. They shouldn't be setting `CC`.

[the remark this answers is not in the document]
jakirkham: Besides what `go` is doing will break things for us [in other places]( https://github.com/conda-forge/nvcc-feedstock/blob/5e21ee5bf0c8f2bf5829c21521f48c3cff8bc419/recipe/install_nvcc.sh#L53 ).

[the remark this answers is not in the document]
pearu: Right. But at least, the current PR allows a workaround to the `go` issue for the `nvcc` script (one can reset `$CXX` before calling `nvcc`).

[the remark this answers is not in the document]
pearu: Btw, any help convincing the go feedstock maintainers about this issue is appreciated.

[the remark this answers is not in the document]
jakirkham: The [last comment]( https://github.com/conda-forge/go-feedstock/issues/47#issuecomment-512681397 ) seems to suggest they are open to changing things, but lack the bandwidth to do so.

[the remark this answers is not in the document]
jakirkham: If we are to come up with a workaround, I'd be more inclined to see it live in a recipe that use `go` and `nvcc` than in the `nvcc` package.

recipe/install_nvcc.sh
kkraus14: Because once we activate the environment there's a new `nvcc` higher in the path which is our shim script to call into the actual `nvcc` 😄

jakirkham: > Because once we activate the environment there's a new `nvcc` higher in the path which is our shim script to call into the actual `nvcc` 😄

Right! 🤦‍♂

recipe/install_nvcc.sh
kkraus14: This is a check that the directory exists, a user can specify an arbitrary value for `CUDA_HOME` so it may not exist as an actual directory.

recipe/install_nvcc.sh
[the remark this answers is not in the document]
kkraus14: Agree with @jakirkham here that workarounds should ideally be at the intersection of `nvcc` and `go` until we can get this resolved upstream in `go`

recipe/install_nvcc.sh
kkraus14: I think it's just my code formatter doing this, will revert.

recipe/install_nvcc.sh
jakirkham: Ah ok. Thanks for clarifying.

pearu:
The `CONDA_ENV_SYSROOT` suggestion appears to be not needed when using `compiler('cgo')` instead of `go` in dependencies. So, LGTM as well.

recipe/install_nvcc.sh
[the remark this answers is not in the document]
pearu: The change is not required when using `compiler('cgo')`.

jakirkham:
@isuruf, would you have a chance to give this another look?

kkraus14:
Thanks for spending the time to review @pearu, appreciated!

jakirkham:
Thanks @kkraus14! Also thanks everyone for the reviews. 😄

If there was anything missed, please raise an issue and we can follow up separately. Thanks.

jakirkham:
Trying this out in PR ( https://github.com/conda-forge/ucx-split-feedstock/pull/33 ).

PR #26, opened 2019-10-26
Search `/usr/include` for CUDA 10.1 headers

jakirkham:
As CUDA 10.1 moves some headers and libraries from `/usr/local/cuda` to `/usr`, we need to adjust our search strategy for finding them. Since `cudatoolkit` already contains the libraries and is available during the build, this is less of an issue. However we still need to point the compiler to the headers in `/usr/include`. Thus we add `/usr/include` to our compiler flags. To avoid clobbering other search paths we care about, we make sure to add `/usr/include` last. As we are not searching `/usr/lib64`, this shouldn't result in any accidental linkages against system libraries.

ref: https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html

Checklist

### DIFF
--- a/recipe/install_nvcc.sh
+++ b/recipe/install_nvcc.sh
@@ -42,9 +42,9 @@ if [[ \$(grep -q "CUDA Version ${PKG_VERSION}" \${CUDA_HOME}/version.txt) -ne 0 
 fi
 
 export CUDA_HOME="\${CUDA_HOME}"
-export CFLAGS="\${CFLAGS} -I\${CUDA_HOME}/include"
-export CPPFLAGS="\${CPPFLAGS} -I\${CUDA_HOME}/include"
-export CXXFLAGS="\${CXXFLAGS} -I\${CUDA_HOME}/include"
+export CFLAGS="\${CFLAGS} -I\${CUDA_HOME}/include -I/usr/include"
+export CPPFLAGS="\${CPPFLAGS} -I\${CUDA_HOME}/include -I/usr/include"
+export CXXFLAGS="\${CXXFLAGS} -I\${CUDA_HOME}/include -I/usr/include"
 
 # Add \$(libcuda.so) shared object stub to the compiler sysroot.
 # Needed for things that want to link to \$(libcuda.so).
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "nvcc" %}
-{% set number = 4 %}
+{% set number = 5 %}
 
 package:
   name: "{{ name }}"


jakirkham:
@conda-forge-admin, please re-render.

recipe/install_nvcc.sh
```diff
@@ -42,9 +42,9 @@ if [[ \$(grep -q "CUDA Version ${PKG_VERSION}" \${CUDA_HOME}/version.txt) -ne 0
 fi
 
 export CUDA_HOME="\${CUDA_HOME}"
-export CFLAGS="\${CFLAGS} -I\${CUDA_HOME}/include"
-export CPPFLAGS="\${CPPFLAGS} -I\${CUDA_HOME}/include"
-export CXXFLAGS="\${CXXFLAGS} -I\${CUDA_HOME}/include"
+export CFLAGS="\${CFLAGS} -I\${CUDA_HOME}/include -I/usr/include"
```
isuruf: Mixing `-I/usr/include` and the sysroot is asking for trouble. cc @mingwandroid

jakirkham:
For context, here's one case (cuBLAS) where libraries and headers are moved out of `${CUDA_HOME}`.

> With this release, on Linux systems, the cuBLAS libraries listed below are now installed in the `/usr/lib/<arch>-linux-gnu/` or `/usr/lib64/` directories as shared and static libraries. Their interfaces are available in the `/usr/include` directory

ref: https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#cublas-new-features

recipe/install_nvcc.sh
jakirkham: Admittedly I don't like this either. That said, this is just the headers and we are adding them last after those included in the build and host prefixes.

isuruf: No, you ar`e not. Host prefix is included as `-isystem` which has lower priority than `-I`, therefore you need to make this `-isystem`  instead to make it work. But with `-isystem` if you are adding any folder in a sysroot like folder then you need to add all of the folders including the C++ headers folder, etc because `-isystem` is weird like that. This means you won't be able to use the C++ headers from the new compiler (7.3.0) and will be stuck with 4.8.5

jakirkham: Well we could add `-I${PREFIX}/include` and similar here.

isuruf: Even still you are messing with the system includes.

There are 2 options,
1. Symlink all headers cuda needs to a different directory and add it.
2. Use cudatoolkit-dev which has the correct directory structure.

pearu:
The idea of this PR sounds (really) bad as it
(i) contradicts with the idea of keeping system and conda environments separate
(ii) will certainly cause issues for those who wish to switch between different CUDA versions according to this `nvcc` recipe.

Perhaps this issue should be raised upstream as with 10.1 policy, one cannot use two CUDA 10.1+ installations in one system anymore (when installing CUDA libraries to `/usr/{lib,include}`). For instance, the target location such as `/usr/lib/x86_64-linux-gnu/libcublas.so` will be coinciding for the two installations.

On the other hand, the only way to install two or more CUDA 10.1+ toolkits to one system without conflicts is to use cuda toolkit installer `--installpath` option:
```
$ bash cuda_10.1.243_418.87.00_linux.run --help
...
  --librarypath=<path>
    Install libraries to the <path> directory. If this flag is not provided,
    the default path of your distribution is used. This flag only applies to
    libraries installed outside of the CUDA Toolkit path.

  --installpath=<path>
    Install everything to the <path> directory. This flag sets the same values
    as the toolkitpath, samplespath, and librarypath options.
...
```

There are two kinds of `nvcc` recipe users:
1. the ones who always use the latest CUDA toolkit and this PR would resolve the cuBlas issue (or equivalently, (ii) would not be an issue) but the issue (i) will remain.
2. the ones who use different versions of CUDA toolkit via specifying `--installpath` and this PR would not be needed but the issue (i) remains.

In conclusion, I see two options:
1. reject this PR as it is and recommend using `--installpath` for all `nvcc` recipe users.
2. adjust the PR so that using `/usr` would not be enabled by default.

recipe/install_nvcc.sh
jakirkham: > Symlink all headers cuda needs to a different directory and add it.

Reorganizing the Docker image a little bit could make sense. However it does seem like we are going in the opposite direction of CUDA Toolkit development, which is a concern.

> Use cudatoolkit-dev which has the correct directory structure.

Suspect it will have the same issue as the installer itself is placing these files in `/usr`.

jakirkham:
Yep, I hear what you are saying and have similar concerns. FWIW this issue is also being raised at NVIDIA.

That said, this does not change the fact that we have this issue today and we need to address it somehow. So the question now becomes what do we do to handle this issue.

> On the other hand, the only way to install two or more CUDA 10.1+ toolkits to one system without conflicts is to use cuda toolkit installer `--installpath` option

Have you tried this out? If so, are you seeing cuBLAS, etc. showing up in the intended install location or are they still winding up in `/usr`?

pearu:
Yes, in my box I have installed 9.2, 10.0, and 10.1 versions to /usr/local/cuda-X.Y.Z/ using `--installpath` and cuBLAS libraries and include files are under the installpath directory.
Here's the cudatoolkit installation commands that I have used for 10.1, for instance:
```
sudo bash cuda_10.1.243_418.87.00_linux.run \
  --toolkit --toolkitpath=/usr/local/cuda-10.1.243/ \
  --installpath=/usr/local/cuda-10.1.243/ --override \
  --no-opengl-libs --no-man-page --no-drm --silent
```
Here's the result:
```
$ ls /usr/local/cuda-10.1.243/lib64/libcublas*
/usr/local/cuda-10.1.243/lib64/libcublas.so
/usr/local/cuda-10.1.243/lib64/libcublas.so.10
/usr/local/cuda-10.1.243/lib64/libcublas.so.10.2.1.243
/usr/local/cuda-10.1.243/lib64/libcublasLt.so
/usr/local/cuda-10.1.243/lib64/libcublasLt.so.10
/usr/local/cuda-10.1.243/lib64/libcublasLt.so.10.2.1.243
/usr/local/cuda-10.1.243/lib64/libcublasLt_static.a
/usr/local/cuda-10.1.243/lib64/libcublas_static.a
```

jakirkham:
This is now handled in the Docker image ( https://github.com/conda-forge/docker-images/pull/134 ). Thanks @bdice! 😄

PR #33, opened 2019-11-16
Add -ICONDA_PREFIX/include to nvcc

pearu:
Checklist

This PR adds `-I$CONDA_PREFIX/include` to `nvcc` options to ease finding the library headers installed by conda.

### DIFF
--- a/recipe/install_nvcc.sh
+++ b/recipe/install_nvcc.sh
@@ -127,6 +127,6 @@ EOF
 mkdir -p "${PREFIX}/bin"
 cat > "${PREFIX}/bin/nvcc" <<EOF
 #!/bin/bash
-"\${CUDA_HOME}/bin/nvcc" -ccbin "\${CXX}" \$@
+"\${CUDA_HOME}/bin/nvcc" -ccbin "\${CXX}" -I\${PREFIX}/include \$@
 EOF
 chmod +x "${PREFIX}/bin/nvcc"
--- a/recipe/install_nvcc.sh
+++ b/recipe/install_nvcc.sh
@@ -127,6 +127,6 @@ EOF
 mkdir -p "${PREFIX}/bin"
 cat > "${PREFIX}/bin/nvcc" <<EOF
 #!/bin/bash
-"\${CUDA_HOME}/bin/nvcc" -ccbin "\${CXX}" -I\${PREFIX}/include \$@
+"\${CUDA_HOME}/bin/nvcc" -ccbin "\${CXX}" -I\${CONDA_PREFIX}/include \$@
 EOF
 chmod +x "${PREFIX}/bin/nvcc"
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "nvcc" %}
-{% set number = 5 %}
+{% set number = 6 %}
 
 package:
   name: "{{ name }}"


pearu:

recipe/install_nvcc.sh
```diff
@@ -127,6 +127,6 @@ EOF
 mkdir -p "${PREFIX}/bin"
 cat > "${PREFIX}/bin/nvcc" <<EOF
 #!/bin/bash
-"\${CUDA_HOME}/bin/nvcc" -ccbin "\${CXX}" \$@
+"\${CUDA_HOME}/bin/nvcc" -ccbin "\${CXX}" -I\${CONDA_PREFIX}/include \$@
```
isuruf: This points to the build prefix in conda-build.

pearu: Notice the backslash: `\${CONDA_PREFIX}`

isuruf: and ?

isuruf: Are you intending to add `$BUILD_PREFIX/include` when this package is used as a compiler in conda-build?

pearu: Yes, the application is about using nvcc inside a conda-build of another recipe.
IIUC, the `include` prefix should be the same as the prefix of `$CXX`.

isuruf: That's wrong. `BUILD_PREFIX` (`CONDA_PREFIX`) is for build tools in the build system. like the compiler, cmake, make etc. `PREFIX` is for the host tools where the libraries are located.

pearu: So, when nvcc is run in conda-build, the prefix should be `\${PREFIX}`.
And when nvcc is run in user's conda environment, the prefix should be `\${CONDA_PREFIX}`.
Is this correct?
If not, my next guess would be to use `${PREFIX}` (the conda prefix transformations are still confusing to me).

kkraus14: Considering `nvcc` is a compiler and we likely will install it into the `BUILD_PREFIX` should we include the `BUILD_PREFIX` first (if it exists) and then include `PREFIX`?

jakirkham: I think what @pearu [said above]( https://github.com/conda-forge/nvcc-feedstock/pull/33#discussion_r347106624 ) was correct. Namely `PREFIX` in the build and `CONDA_PREFIX` outside of a build.

Should add we can check whether `CONDA_BUILD` is set to `1` to determine if we are in a conda-build context.

ref: https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#environment-variables-set-during-the-build-process

jakirkham: @isuruf, is there a way we could bake the include path into [ctng-compiler-activation]( https://github.com/conda-forge/ctng-compiler-activation-feedstock )?

isuruf:
The build system of a package should use `CFLAGS, CPPFLAGS, CXXFLAGS, LDFLAGS`. Adding this here would only just mask the issue that these env variable are not used. `-1` on this PR.

pearu:
Apparently, `nvcc` does not use `CFLAGS`, `CPPFLAGS`, `CXXFLAGS`, `LDFLAGS`.
For instance, running `nvcc -v test.cu` (that has `#include "boost/regexp.hpp"`) in test conda environment with `boost-cpp` installed, the output is:
```
#$ _SPACE_= 
#$ _CUDART_=cudart
#$ _HERE_=/usr/local/cuda-10.1.243/bin
#$ _THERE_=/usr/local/cuda-10.1.243/bin
#$ _TARGET_SIZE_=
#$ _TARGET_DIR_=
#$ _TARGET_DIR_=targets/x86_64-linux
#$ TOP=/usr/local/cuda-10.1.243/bin/..
#$ NVVMIR_LIBRARY_DIR=/usr/local/cuda-10.1.243/bin/../nvvm/libdevice
#$ LD_LIBRARY_PATH=/usr/local/cuda-10.1.243/bin/../lib:/usr/local/cuda-10.1.243/lib64:
#$ PATH=/usr/local/cuda-10.1.243/bin/../nvvm/bin:/usr/local/cuda-10.1.243/bin:/home/pearu/miniconda3/conda-bld/nvcc_1574067209285/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_p:/home/pearu/miniconda3/conda-bld/nvcc_1574067209285/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_p/bin:/home/pearu/miniconda3/conda-bld/nvcc_1574067209285/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_p/bin:/usr/local/cuda-10.1.243/bin:/home/pearu/miniconda3/bin:/home/pearu/miniconda3/condabin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
#$ INCLUDES="-I/usr/local/cuda-10.1.243/bin/../targets/x86_64-linux/include"  
#$ LIBRARIES=  "-L/usr/local/cuda-10.1.243/bin/../targets/x86_64-linux/lib/stubs" "-L/usr/local/cuda-10.1.243/bin/../targets/x86_64-linux/lib"
#$ CUDAFE_FLAGS=
#$ PTXAS_FLAGS=
#$ "/home/pearu/miniconda3/conda-bld/nvcc_1574067209285/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_p/bin"/x86_64-conda_cos6-linux-gnu-c++ -D__CUDA_ARCH__=300 -E -x c++  -DCUDA_DOUBLE_MATH_FUNCTIONS -D__CUDACC__ -D__NVCC__  "-I/usr/local/cuda-10.1.243/bin/../targets/x86_64-linux/include"    -D__CUDACC_VER_MAJOR__=10 -D__CUDACC_VER_MINOR__=1 -D__CUDACC_VER_BUILD__=243 -include "cuda_runtime.h" -m64 "test.cu" > "/tmp/tmpxft_00002366_00000000-8_test.cpp1.ii" 
test.cu:2:10: fatal error: boost/regex.hpp: No such file or directory
 #include "boost/regex.hpp"
          ^~~~~~~~~~~~~~~~~
compilation terminated.
```
Note that `CXXFLAGS` contains the required `-I$PREFIX/include` and `$PREFIX/include/boost/` exists:
```
CXXFLAGS=-fvisibility-inlines-hidden -std=c++17 -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -I$PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/nvcc_linux-64-10.1 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix -I/usr/local/cuda-10.1.243/include
```

pearu:
`nvcc` seems to use `${CUDA_HOME}/bin/nvcc.profile` for setting the required paths. A typical `nvcc.profile` contains
```
TOP              = $(_HERE_)/..
NVVMIR_LIBRARY_DIR = $(TOP)/nvvm/libdevice
LD_LIBRARY_PATH += $(TOP)/lib:
PATH            += $(TOP)/nvvm/bin:$(_HERE_):
INCLUDES        +=  "-I$(TOP)/$(_TARGET_DIR_)/include" $(_SPACE_)
LIBRARIES        =+ $(_SPACE_) "-L$(TOP)/$(_TARGET_DIR_)/lib$(_TARGET_SIZE_)/stubs" "-L$(TOP)/$(_TARGET_DIR_)/lib$(_TARGET_SIZE_)"
CUDAFE_FLAGS    +=
PTXAS_FLAGS     +=
```
I don't know yet if one can have a customized `nvcc.profile` that could be defined for conda environment. For instance, `nvcc` does not have an option to specify the path to `nvcc.profile`.

pearu:
@isuruf , it seems to me that the set of environment flags `CFLAGS`, `CPPFLAGS`, `CXXFLAGS`, `LDFLAGS` is not sufficient for specifying include/library paths for `nvcc` compilation process.
The C++ compiler specified in `-ccbin` (and for which `CXXFLAGS` is effective) is used later in the process but that will not help here as `nvcc` calls CPP on `.cu` file earlier.

So I think your "-1 on this PR" might be premature because the assumption that `nvcc` should use the mentioned environment flags may be wrong. What do you think?

Another approach would be to use
```
export CPLUS_INCLUDE_PATH=\${PREFIX}/include
```
(tested) but that would be equivalent to using `-I` option except that it would affect other compilers as well.

isuruf:
> Apparently, nvcc does not use CFLAGS, CPPFLAGS, CXXFLAGS, LDFLAGS.

`gcc` and `clang` doesn't use these either. It's the job of the build system for a package to pass these to the compiler and linker.

pearu:
OK, I see your point, @isuruf .

pearu:
Closing as the build system or the user has to specify the proper include directory options.
For instance, when having `test.cpp`
```
#include <boost/regex.hpp>
int main() {
  return 0;
}
```
then all the following commands fail:
```
nvcc test.cpp
g++ test.cpp
clang++ test.cpp 
```
while all the following commands succeed:
```
nvcc -I$CONDA_PREFIX/include test.cpp
g++ -I$CONDA_PREFIX/include test.cpp
clang++ -I$CONDA_PREFIX/include test.cpp 
```

PR #43, opened 2020-09-23
Add CUDA 11.0

jakirkham:
Manually adds in `conda_build_config.yaml` with CUDA 11.0.

Checklist


recipe/meta.yaml
```diff
@@ -20,6 +20,8 @@ outputs:
         strong:
           - cudatoolkit {{ cuda_compiler_version }}|{{ cuda_compiler_version }}.*
     requirements:
+      build:
+        - sysroot_linux-64 2.17  # [cuda_compiler_version == "11.0"]
```
jakirkham, line 24: @beckermr, does it make sense to have this as a `run` dependency or maybe `run_exports`? Am wondering how we can ensure packages downstream from `nvcc` use GLIBC 2.17 if they are using CUDA 11.0.

jakirkham:
> * License is not an SPDX identifier (or a custom LicenseRef) nor an SPDX license expression.

Addressing in PR ( https://github.com/conda-forge/nvcc-feedstock/pull/44 ).

recipe/meta.yaml
```diff
@@ -20,6 +20,8 @@ outputs:
         strong:
           - cudatoolkit {{ cuda_compiler_version }}|{{ cuda_compiler_version }}.*
```
isuruf, line 21:
```suggestion
          - cudatoolkit {{ cuda_compiler_version }}|{{ cuda_compiler_version }}.*
          - __glibc >=2.17   # [linux]
```

recipe/meta.yaml
isuruf: Yes, add this to `run`. You don't need it in `build`

jakirkham: Wouldn't we want it in both to make sure that conda-build flags this as requiring GLIBC 2.17?

recipe/meta.yaml
jakirkham: Well we only need this if CUDA 11.0 is being used. Should that be in the selector as well then?

isuruf: yes

recipe/meta.yaml
isuruf: why would you need the sysroot package to build the `nvcc` package?

jakirkham: We [use the C compiler]( https://github.com/conda-forge/nvcc-feedstock/blob/f2ee3a4798aaaa95680f55113026dd7a06f69478/recipe/meta.yaml#L24-L25 ) to configure some things during the build. Maybe this should be in `host` and `run` instead of `build`?

beckermr: This package itself provides a compiler, right?

If so then it should have an export on `__glibc >=2.17` as @isuruf added below. 

If it is a compiler itself, then when in use it def needs the sysroot in run. 

The others are more subtle. It does not hurt to have it in build, so if you are worried you might add it.

jakirkham:
Also cc-ing @scopatz for vis 🙂

recipe/meta.yaml
jakirkham: Sure no objections to the suggestions thus far. Just trying to make sure we have all of our bases covered 🙂

Part of the concern with `build` and `host` is we are [installing stuff in the compiler's sysroot]( https://github.com/conda-forge/nvcc-feedstock/blob/4712ce14ce41b7ce7d9f20d092c98392bd70eb6d/recipe/install_nvcc.sh#L77 ). So want to make sure we handle that correctly.

beckermr: you probably want the sysroot package in host then? you may want to ignore some of the run exports depending what the nvcc compiler needs in terms of glibc.

jakirkham: Ok I think I've addressed this in the new commits. Please let me know what you think.

recipe/meta.yaml
jakirkham: Ok this should be fixed now. Please take another look.

jakirkham:
Any thoughts on this @isuruf @beckermr? 🙂

beckermr:
Can we inspect the output package for the `__glibc >=2.17` constraint? IDK if we want it to have one if it does. Does the compiler need cos7 or just the things it builds?

jakirkham:
CUDA 11.0 is only supported for GLIBC 2.17+. So was trying to capture that somehow. Isuru recommended this change. Should we handle this differently? Both as the Docker image it uses is CentOS 7 based and as a result the binaries it produced will require GLIBC 2.17+.

Happy for us to inspect the package further. How should we do that? Is there a way to upload a draft of the package on CI?

beckermr:
Sorry, let me be more clear. To run the nvcc package produced here, do I need glibc >=2.17?

jakirkham:
For CUDA 11.0, yes.

jakirkham:
Just to provide more context, this package acts as a shim to interact with an actual install of `nvcc`. So all `nvcc` installs for CUDA 11.0 require GLIBC 2.17+. I suppose this package could be installed in an environment without that, but it wouldn't be of much practical use.

recipe/meta.yaml
```diff
@@ -19,16 +19,20 @@ outputs:
...
+          - __glibc >=2.17  # [cuda_compiler_version == "11.0"]
     requirements:
       host:
         # Needed to symlink libcuda into sysroot libs.
         - {{ compiler("c") }}
+        - sysroot_linux-64 2.17  # [cuda_compiler_version == "11.0"]
       run:
+        - sysroot_linux-64 2.17  # [cuda_compiler_version == "11.0"]
```
beckermr, line 29: If you need glibc >=2.17 at runtime, then we need to make sure the virtual package constraint makes it here.

recipe/conda_build_config.yaml
```diff
@@ -0,0 +1,19 @@
...
+docker_image:                                   # [linux]
+  - condaforge/linux-anvil-comp7                # [linux64]
+  - condaforge/linux-anvil-cuda:9.2             # [linux64]
+  - condaforge/linux-anvil-cuda:10.0            # [linux64]
+  - condaforge/linux-anvil-cuda:10.1            # [linux64]
+  - condaforge/linux-anvil-cuda:10.2            # [linux64]
+  - condaforge/linux-anvil-cuda:11.0            # [linux64]
+
+  - condaforge/linux-anvil-aarch64              # [aarch64]
+  - condaforge/linux-anvil-ppc64le              # [ppc64le]
```
scopatz, line 18: Aren't there different ppc64le cuda versions?

recipe/meta.yaml
isuruf: It's not needed at runtime. Only at runtime of downstream packages.

jakirkham: Well downstream packages should be built with GLIBC 2.17. So wouldn't we want that constraint to make sure errors are raised if that is not met?

beckermr: @isuruf @jakirkham wrote this

> Just to provide more context, this package acts as a shim to interact with an actual install of nvcc. So all nvcc installs for CUDA 11.0 require GLIBC 2.17+. I suppose this package could be installed in an environment without that, but it wouldn't be of much practical use.

I think GLIBC 2.17+ is needed at runtime according to this statement.

isuruf: This is a shim package. there's no need to have glibc 2.17 for a shim package with just an activation script.

recipe/conda_build_config.yaml
jakirkham: I took these from [pinnings]( https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/53a32332a1553eb37b3edf4bf96b3311314f2fba/recipe/conda_build_config.yaml#L126-L127 ). Though they are [not being used yet]( https://github.com/conda-forge/nvcc-feedstock/blob/f2ee3a4798aaaa95680f55113026dd7a06f69478/recipe/meta.yaml#L9 ).

Admittedly we might want to add builds for other architectures, but I would rather save that for a future PR.

recipe/meta.yaml
beckermr: Fair but there are two things here. The second is

> I suppose this package could be installed in an environment without that, but it wouldn't be of much practical use.

Why let people install a shim of no use in an env? For the shim to be usable at all for CUDA 11, it needs a machine with GLIBC >=2.17 so it should express that in its constraints.

isuruf: It is usable on a machine without glibc>=2.17 if you use a different compiler like `clang` although there aren't use cases in conda-forge right now.

beckermr: Ah. So this statement

> So all nvcc installs for CUDA 11.0 require GLIBC 2.17+.

Is not true then. Thanks!

isuruf: To run `nvcc`, it is required, but this package does more than that and adds flags to CUDA_HOME, etc.

beckermr: Ahh you mean run clang instead of nvcc. This package is meant for nvcc only as I understand it. It seems like if we want to support running clang instead of nvcc, we should separate the flags from the nvcc stuff and use proper constraints on each one.

recipe/conda_build_config.yaml
scopatz: Yeah, waiting makes sense

recipe/meta.yaml
jakirkham: A little unclear on whether we want this change or not. Went ahead and included it in the latest commit. Though feel free to comment on it if you think changes are needed.

isuruf: I think @beckermr was okay with not having having this change as he marked the conversation as resolved.

beckermr: I don't think it is a great idea to leave this package with formally incorrect requirements so it can be hacked on, but I  also don't really care. I was asking questions to make sure I understood what was going on.

jakirkham: Maybe we can comment on the change more directly? It's in the latest commit below.

recipe/meta.yaml
```diff
@@ -19,16 +19,22 @@ outputs:
...
+          - __glibc >=2.17  # [cuda_compiler_version == "11.0"]
     requirements:
       host:
         # Needed to symlink libcuda into sysroot libs.
         - {{ compiler("c") }}
+        - sysroot_linux-64 2.17  # [cuda_compiler_version == "11.0"]
+        - __glibc >=2.17         # [cuda_compiler_version == "11.0"]
       run:
+        - sysroot_linux-64 2.17  # [cuda_compiler_version == "11.0"]
+        - __glibc >=2.17         # [cuda_compiler_version == "11.0"]
```
jakirkham, line 31: Does this address your concerns, @beckermr?

beckermr: It does since this is required to actually use the package in that case. However, @isuruf didn't want this constraint since he wants to use clang w/ some of the stuff in this package but not all of it. For that case, I would think we'd want a separate cuda-only activation script that the nvcc one would depend on.

beckermr: Again I don't really care. I am trying to make sure I understand what is going on.

jakirkham: Having a separate package could make sense.

That said, CentOS 6 reaches its [EOL 30 Nov 2020]( https://wiki.centos.org/About/Product ). So in a couple months. I'm not sure how much longer we plan to support it, but maybe this becomes a non-issue in the near future.

jakirkham:
@conda-forge-admin, please re-render.

mike-wendt (approved):

recipe/meta.yaml
kkraus14: +1 to having a separate cuda activation script package. From the outside, the point of this package is to get the `nvcc` compiler. This package handling the activation script things is an implementation detail that could change at any point in the future as long as `nvcc` is still being delivered in some way.

jakirkham:
Any more comments on this? 🙂

kkraus14 (approved):

jakirkham:
Thanks all! 😄

If we missed anything, please feel free to raise an issue or PR 🙂

jakirkham:
Switching to a manually added migrator to handle this one instead of hacking in a `conda_build_config.yaml` ( https://github.com/conda-forge/nvcc-feedstock/pull/47 ). Mostly a cosmetic change, but does make things a bit more standardized in the migration.

PR #49, opened 2020-10-31
Add Windows

jaimergp:
Checklist
* [X] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [X] Ensured the license file is being packaged.

## Bringing CUDA  to Windows

We need three things
- `cudatoolkit` packages (submitted by @scopatz)
- `nvcc` metapackage (this PR)
- [X] Install drivers via `conda-forge-ci-setup` -- see https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/129
- [X] Add migrator to `conda-forge-pinning` -- see https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/914
  - Added migrator manually for now

Questions:
- [X] There's no lib stubs on Windows, so I am symlinking / copying `cuda.lib` from `CUDA_HOME\lib\x64` to `LIBRARY_LIB`. Is this ok?
- [X] The env var is `CUDA_PATH`, not `CUDA_HOME`, but I am aliasing this one just in case.
- [X] There's no `cuda-gdb` so I am using `nvcc` as a fallback pathfinder
- [X] CUDA 9.2 requires `vs2015`, which is now deprecated in conda-forge. There _could be_ a [workaround](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/914#discussion_r515657282), but looks tricky.


jaimergp:
This passes locally but the CLI installation of CUDA doesn't modify PATH or sets CUDA_PATH. I'll look into it tomorrow!

jaimergp:
Looks like it's working. I have updated the PR description with comments and questions.

isuruf (changes requested):
Blocked by https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/129

recipe/windows/deactivate.bat
```diff
@@ -0,0 +1,41 @@
...
+
+:: JRG: Should CONDA_BUILD_SYSROOT be replaced with LIBRARY_PREFIX?
+:: set "CONDA_BUILD_SYSROOT=%CONDA_PREFIX%"
+
+:: Remove or restore `cuda.lib` from the compiler sysroot.
+set "LIBCUDA_SO_CONDA_NVCC_BACKUP=%LIBRARY_LIB%\cuda.lib-conda-nvcc-backup"
+if exist "%LIBCUDA_SO_CONDA_NVCC_BACKUP%" (
+    ren "%LIBCUDA_SO_CONDA_NVCC_BACKUP%" "%LIBRARY_LIB%\cuda.lib"
+) else (
+    del "%LIBRARY_LIB%\cuda.lib"
```
isuruf, line 40: Can this fail somehow? If it does and the build passes, we'll be packaging this against its license.

jaimergp: I can add an `errorlevel` check there and also text for non-existence in the `meta.yaml` tests. Would that be enough?

jaimergp:
I think I was a bit too quick to push the new commits. The new `conda-forge-ci-setup` is not deployed yet 😬

jaimergp:
I guess CUDA 11 places their Windows files elsewhere. I'll investigate locally.

recipe/windows/deactivate.bat
jaimergp: See now, together with https://github.com/conda-forge/nvcc-feedstock/pull/49/files#diff-e3a5306c0a02b6cefeced655643e5e6a4625e9c8af54650c829f7931ca44009dR74-R80

isuruf: But that test is only for this recipe. What happens in downstream packages?

How about generating this `cuda.lib` ourselves from the DLL so that we can redistribute it?

jaimergp:
That did it. CUDA 11 does not ship a `%CUDA_HOME%\version.txt` file, so I am querying `%CUDA_HOME%\bin\nvcc.exe --version` for `release %PKG_VERSION%` as a fallback mechanism.

recipe/windows/deactivate.bat
jaimergp: Something [like this](https://stackoverflow.com/questions/9360280/how-to-make-a-lib-file-when-have-a-dll-file-and-a-header-file)?

These are the DLLs available under `\bin`. Which one applies here?

```
Directory: C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v11.0\bin


    Length Name
    ------ ----
  89057280 cublas64_11.dll
 167523328 cublasLt64_11.dll
    335872 cudart32_110.dll
    401408 cudart64_110.dll
 162878976 cufft64_10.dll
    272896 cufftw64_10.dll
   1305600 cuinj64_110.dll
  52446720 curand64_10.dll
 525173248 cusolver64_10.dll
 314601984 cusolverMg64_10.dll
 159789568 cusparse64_11.dll
    232448 nppc64_11.dll
  10907648 nppial64_11.dll
   4569600 nppicc64_11.dll
   7840256 nppidei64_11.dll
  51866112 nppif64_11.dll
  26065920 nppig64_11.dll
   6411264 nppim64_11.dll
  20231168 nppist64_11.dll
    203264 nppisu64_11.dll
   2856960 nppitc64_11.dll
   8873472 npps64_11.dll
    299520 nvblas64_11.dll
   3133952 nvjpeg64_11.dll
   5118976 nvrtc-builtins64_110.dll
  16161792 nvrtc64_110_0.dll
```

isuruf: > Something like this?

Yes

> These are the DLLs available under \bin. Which one applies here?

Have a look at `cuda.lib` (It's a text file) It should tell you

jaimergp: Thanks! Had to use `dumpbin /ALL cuda.lib` to figure it out. FIle is `nvcuda.dll`, which should be part of the drivers at `C:\Windows\system32`, but I can't find it in my system. [A separate installation process for the drivers](https://forums.developer.nvidia.com/t/cuda-rc-11-0-installs-a-nvcuda-dll-that-does-not-have-version-information/127865) seems to fix it.

jaimergp: I added a stub now. Tried to generate off the drivers installation in the CI, but `nvcuda.dll` is not present there (I guess because no GPU). So I generated one from my 456.71 Windows installation and copied it over. Is that ok? I left a `generate_stub.bat` script in case it needs to be regenerated again in the future (drivers update?).

isuruf: the file might get outdated in that case. How about this? create an empty `cuda.lib` in `bld.bat` here so that the empty file will get packaged and then overriding that in a `activate.bat`? This way, conda will not package the actual `cuda.lib` in a downstream package because that file is already there.

jaimergp: Oh, that could work, yes. Check last commit 36ade89.

jaimergp:
Anything else needed?

recipe/windows/activate.bat
```diff
@@ -0,0 +1,80 @@
...
+
+:: Add `cuda.lib` shared object stub to the compiler sysroot.
+:: Needed for things that want to link to `cuda.lib`.
+:: Stub is used to avoid getting driver code linked into binaries.
+
+:: Make a backup of `cuda.lib` if it exists -- we make sure this is the case in install_nvcc.bat
+if exist %LIBRARY_LIB%\cuda.lib (
+    set "LIBCUDA_SO_CONDA_NVCC_BACKUP=%LIBRARY_LIB%\cuda.lib-conda-nvcc-backup"
+    ren "%LIBRARY_LIB%\cuda.lib" "%LIBCUDA_SO_CONDA_NVCC_BACKUP%"
+)
```
isuruf, line 71: How about just delete this instead of keeping a backup. There's no need to restore it.

jaimergp: This check is also present in the Linux version. I guess it's trying to prevent we mess with another installation that might be providing that file?

isuruf (approved):

jaimergp:
Thanks! 🥳

jakirkham:
Thanks for all of this work! 😄

PR #50, opened 2020-11-22
Add CUDA 11.1 migrator

jaimergp:
Checklist

Blocked by https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/131 and https://github.com/conda-forge/docker-images/pull/160

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "nvcc" %}
-{% set number = 8 %}
+{% set number = 9 %}
 
 package:
   name: "{{ name }}"
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -20,22 +20,22 @@ outputs:
       run_exports:
         strong:
           - cudatoolkit {{ cuda_compiler_version }}|{{ cuda_compiler_version }}.*
-          - __glibc >=2.17  # [linux and int(cuda_compiler_version.split(".")[0]) >= 11)]
+          - __glibc >=2.17  # [linux and int(cuda_compiler_version.split(".")[0]) >= 11]
     requirements:
       host:
         # Needed to symlink libcuda into sysroot libs.
         - {{ compiler("c") }}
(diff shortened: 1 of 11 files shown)


jaimergp:
That doesn't work even if tests are successful. CUDA 11 is back to 2.12 sysroots.

jaimergp:
Ok, using `cdt_name` does work. I guess that's more sustainable in the future (and makes more sense since 2.17 is tied to cos7?). If this is not ok, we need to figure out a clean way to state that CUDA 11.0+ requires glibc 2.17+...

recipe/meta.yaml
```diff
@@ -20,22 +20,22 @@ outputs:
       run_exports:
         strong:
           - cudatoolkit {{ cuda_compiler_version }}|{{ cuda_compiler_version }}.*
-          - __glibc >=2.17  # [linux and cuda_compiler_version == "11.0"]
+          - __glibc >=2.17  # [linux and cuda_compiler_version in ("11.0", "11.1")]
```
kkraus14, line 23: Can we use string comparisons so we don't have to keep adding here whenever a new CUDA version is released?
```suggestion
          - __glibc >=2.17  # [linux and cuda_compiler_version >= "11.0"]
```

jaimergp: I tried with int-tuples but the selector syntax does not work well with them. See previous commits and maybe you'll see something I missed?

jaimergp: Hm, I figured out a way to do it but it's _not_ pretty. See commit 7ca2d6a.

kkraus14: Do we need to convert the major version to an int? I was suggesting using string comparison, but that does fail if/when there's a CUDA version 100 😄.

jaimergp: 9.2 would be greater than 11 if we use string comparison:

```
>>> "9.2" > "11.0"
True
```

kkraus14: Resolving this, forgot we're still supporting 9.2 for now.

jaimergp:
I'll probably get back to this soon. Looks like all the missing pieces are available now!

PR #52, opened 2020-11-26
[Windows] Freeze PKG_VERSION and add fallback lookup in default installation path

jaimergp:
Checklist

# Part 1

This fixes a subtle bug I just found for the activation scripts on Windows. Since we are not creating the activation scripts from the install script, but by copying them from files, environment variables are not rendered to the final file. Instead the take the value set in the environment when they are executed. For our tests that didn't matter because they got rendered correctly, but when used in other recipes, they might take the wrong value.

For example, the CUDA version check uses `PKG_VERSION`, which will be set for other packages during build time. [Example here](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=245139&view=logs&j=84cf05f7-5267-5f38-137b-63208e75d05b&t=7c51bcc0-b6e0-5e4f-1968-7bb191e490bd&l=348). That `7.5.0` is OpenMM's version, not CUDA!

To fix this, I am rendering `PKG_VERSION` by hand, using `sed` 😬 

This is not producing errors because of the fallback mechanisms for finding `CUDA_HOME`, but that doesn't mean it should not be fixed. Especially if this package is used outside the context of conda-forge.

# Part 2

For some reason, in some feedstocks `conda-forge-ci-setup` mechanisms do not reach the recipe (e.g. `CUDA_PATH` is not set and `PATH` does not contain `nvcc.exe`). See #53 for more details.

As a workaround I am adding a fallback lookup in the default installation path, where the activation script might find nvcc.exe.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "nvcc" %}
-{% set number = 7 %}
+{% set number = 8 %}
 
 package:
   name: "{{ name }}"
@@ -22,6 +22,8 @@ outputs:
           - cudatoolkit {{ cuda_compiler_version }}|{{ cuda_compiler_version }}.*
           - __glibc >=2.17  # [linux and cuda_compiler_version == "11.0"]
     requirements:
+      build:      # [win]
+        - m2-sed  # [win]
       host:
         # Needed to symlink libcuda into sysroot libs.
         - {{ compiler("c") }}
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "nvcc" %}
-{% set number = 8 %}
+{% set number = 9 %}
 
 package:
   name: "{{ name }}"
@@ -65,3 +65,4 @@ extra:
     - kkraus14
     - mike-wendt
     - raydouglass
+    - jaimergp
--- a/recipe/windows/activate.bat
+++ b/recipe/windows/activate.bat
@@ -22,12 +22,24 @@ if defined CXXFLAGS (
     set "CXXFLAGS_CONDA_NVCC_BACKUP=%CXXFLAGS%"
 )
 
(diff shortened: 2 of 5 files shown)


jaimergp:
@conda-forge/nvcc I think this is ready for review. If you agree, I'd like to add me as a maintainer for the Windows bits too.

kkraus14:
> @conda-forge/nvcc I think this is ready for review. If you agree, I'd like to add me as a maintainer for the Windows bits too.

Sounds good to me, please add yourself as a maintainer on the recipe!

recipe/windows/activate.bat
```diff
@@ -22,12 +22,24 @@ if defined CXXFLAGS (
     set "CXXFLAGS_CONDA_NVCC_BACKUP=%CXXFLAGS%"
 )
 
-:: Default to using nvcc.exe to specify %CUDA_HOME%
+:: Default to using nvcc.exe to specify %CUDA_PATH%
+:: Things we try:
+:: 1) CUDA_PATH might be defined already in the env, use it in that case
+:: 2) Try to locate nvcc.exe in PATH and calculate CUDA_PATH from there (two levels up)
+:: 3) Look if nvcc.exe can be foudn in the default CUDA installation path and calculate CUDA_PATH from there (two levels up)
```
kkraus14, line 29:
```suggestion
:: 3) Look if nvcc.exe can be found in the default CUDA installation path and calculate CUDA_PATH from there (two levels up)
```

kkraus14 (approved):
One typo, but approach LGTM

recipe/meta.yaml
```diff
@@ -1,5 +1,5 @@
 {% set name = "nvcc" %}
-{% set number = 7 %}
+{% set number = 8 %}
```
kkraus14, line 2: Another PR just merged where I think we need to bump this to 9 now.

jaimergp:
Thanks!

Issue #34, opened 2019-12-19
[BUG] Redefinition of '-ccbin' argument


teju85:
Issue:
If the project build system explicitly tries to customize compiler path by passing `-ccbin` option, then this package ends up with the following error:
`nvcc fatal   : redefinition of argument 'compiler-bindir'`
This seems to happen because the nvcc wrapper script explicitly tries to pass `-ccbin` itself.
<br/>

Environment (<code>conda list</code>):
[conda-list.txt](https://github.com/conda-forge/nvcc-feedstock/files/3982456/conda-list.txt)
<br/>

Details about  <code>conda</code> and system ( <code>conda info</code> ):
[conda-info.txt](https://github.com/conda-forge/nvcc-feedstock/files/3982465/conda-info.txt)

Additional information:
To know more about why we pass `-ccbin` option ourselves, refer to our [cuML CMakeLists.txt](https://github.com/rapidsai/cuml/blob/branch-0.12/cpp/CMakeLists.txt#L172)

carterbox:
I have this same issue, but in my `CMakeLists.txt` I define
```cmake
set(CMAKE_CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER})
```
Otherwise, `nvcc` doesn't always choose the whatever compiler is set using the environment variable `CXX` as the host compiler. I guess this is the same reason that the nvcc wrapper uses the `-ccbin` parameter.

As a work-around, instead of the above CMAKE call, I can also set the environment variable `CUDAHOSTCXX` whenever I'm not using `conda-forge nvcc`.

Issue #51, opened 2020-11-25
nvcc package should constrain c / cxx compiler versions based on cuda version


kkraus14:
Compatibility matrix: https://gist.github.com/ax3l/9489132

Latest versions supported:
| CUDA Version | GCC Version |
| -------------- | ------------- |
| 9.2 | 7 |
| 10.0 | 7 |
| 10.1 | 8 |
| 10.2 | 8 |
| 11.0 | 9 |
| 11.1 | 10 |

jaimergp:
I was thinking about this too. Is this to be addressed as part of the global pinnings at `conda-forge-ci-pinning` via `zipped_keys`?

jakirkham:
IIRC Keith we are only worried about this for packages that actually contain CUDA kernel code right? Not packages that merely link to CUDA libraries. Is that correct or am I missing something?

kkraus14:
Any code that compiled with nvcc, so even if they don't have device code if they use a `.cu` extension it will go through nvcc.

leofang:
I think this should be fixed asap, as Conda-Forge has updated the global pinning to gcc9, and cupy-feedstock starts to see tons of build errors: https://github.com/conda-forge/cupy-feedstock/pull/80#issuecomment-734635067.

leofang:
> Any code that compiled with nvcc, so even if they don't have device code if they use a `.cu` extension it will go through nvcc.

Even worse, anything not compiled by nvcc but includes cuda_runtime.h will be affected (see the cupy discussion linked above).

kkraus14:
Looks like it's more than just libraries with device code based on the above.

jakirkham:
What happens if we build against GCC 7 and the runtime library from GCC 9 is pulled in by the compiled packages?

kkraus14:
> What happens if we build against GCC 7 and the runtime library from GCC 9 is pulled in by the compiled packages?

As far as I'm aware that should be fine, there's just compiler checks in the headers that prevent using too new of a compiler.

leofang:
Update: with https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1052 all CUDA builds fall back to gcc 7 now, so it works for all CUDA versions (9.2 - 11.0). It'd be great to enable gcc 9 for CUDA 11.0+ only.

jakirkham:
[...]

PR #54, opened 2020-12-07
nvcc wrapper: Add -ccbin only if not already supplied

mbargull:
Checklist

Closes gh-34
Probably fixes https://github.com/conda-forge/faiss-split-feedstock/pull/17#issuecomment-739909855
cc @h-vetinari

### DIFF
--- a/recipe/install_nvcc.sh
+++ b/recipe/install_nvcc.sh
@@ -122,13 +122,14 @@ EOF
 
 # Create `nvcc` script in `bin` so it can be easily run.
 mkdir -p "${PREFIX}/bin"
-cat > "${PREFIX}/bin/nvcc" <<EOF
+cat > "${PREFIX}/bin/nvcc" <<'EOF'
 #!/bin/bash
-for i ; do
-  case \$i in -ccbin)
-    exec "\${CUDA_HOME}/bin/nvcc" "\${@}"
+for arg in "${@}" ; do
+  case ${arg} in -ccbin)
+    # If -ccbin argument is already provided, don't add an additional one.
+    exec "${CUDA_HOME}/bin/nvcc" "${@}"
   esac
 done
-exec "\${CUDA_HOME}/bin/nvcc" -ccbin "\${CXX}" "\${@}"
+exec "${CUDA_HOME}/bin/nvcc" -ccbin "${CXX}" "${@}"
 EOF
 chmod +x "${PREFIX}/bin/nvcc"
--- a/recipe/install_nvcc.sh
+++ b/recipe/install_nvcc.sh
@@ -124,6 +124,11 @@ EOF
 mkdir -p "${PREFIX}/bin"
 cat > "${PREFIX}/bin/nvcc" <<EOF
 #!/bin/bash
+for i ; do
+  case \$i in -ccbin)
+    exec "\${CUDA_HOME}/bin/nvcc" "\${@}"
+  esac
+done
 exec "\${CUDA_HOME}/bin/nvcc" -ccbin "\${CXX}" "\${@}"
 EOF
 chmod +x "${PREFIX}/bin/nvcc"
--- a/recipe/install_nvcc.sh
+++ b/recipe/install_nvcc.sh
@@ -124,6 +124,6 @@ EOF
 mkdir -p "${PREFIX}/bin"
 cat > "${PREFIX}/bin/nvcc" <<EOF
 #!/bin/bash
-"\${CUDA_HOME}/bin/nvcc" -ccbin "\${CXX}" \$@
+"\${CUDA_HOME}/bin/nvcc" -ccbin "\${CXX}" "\${@}"
 EOF
 chmod +x "${PREFIX}/bin/nvcc"
--- a/recipe/install_nvcc.sh
+++ b/recipe/install_nvcc.sh
@@ -124,6 +124,6 @@ EOF
 mkdir -p "${PREFIX}/bin"
 cat > "${PREFIX}/bin/nvcc" <<EOF
 #!/bin/bash
-"\${CUDA_HOME}/bin/nvcc" -ccbin "\${CXX}" "\${@}"
+exec "\${CUDA_HOME}/bin/nvcc" -ccbin "\${CXX}" "\${@}"
 EOF
 chmod +x "${PREFIX}/bin/nvcc"
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -24,12 +24,7 @@ outputs:
     requirements:
       host:
         # Needed to symlink libcuda into sysroot libs.
-        # FIXME: nvcc 9.* and 10.* don't work with GCC 9.
-        #        Adding custom GCC versions via conda_build_config.yaml fails
-        #        due to https://github.com/conda/conda-build/issues/3974
-        #        => use hardcoded compiler package names to workaround that.
-        - {{ compiler("c") }}    # [not (linux64 and (cuda_compiler_version or "").startswith(("9.", "10.")))]
-        - gcc_linux-64 7         # [linux64 and (cuda_compiler_version or "").startswith(("9.", "10."))]
+        - {{ compiler("c") }}
         - sysroot_linux-64 2.17  # [linux and cuda_compiler_version == "11.0"]
         - __glibc >=2.17         # [linux and cuda_compiler_version == "11.0"]
       run:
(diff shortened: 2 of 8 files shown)


mbargull:
(NB: If the same is needed on Windows, then I'll leave it for someone else to address. I.e., I don't have time/nerve for Windows CMD hacking.)

jaimergp:
If needed, I'll handle the Windows part once this PR is merged!

h-vetinari:
This is running into the same gcc-vs-nvcc version problem that I had in https://github.com/conda-forge/faiss-split-feedstock/pull/17 resp. https://github.com/conda-forge/faiss-split-feedstock/pull/19:

* nvcc 9.2 requires gcc <= 7
* nvcc 10.x requires gcc <= 8
* nvcc 11.x requires gcc <= 9 [presumably; not _yet_ relevant]

Possible solutions are [manual downgrade](https://github.com/conda-forge/faiss-split-feedstock/pull/18/commits/ecabd2fc678446c13a545497a19e82a92af78ada) (won't survive rerender), [across-the-bank downgrade](https://github.com/conda-forge/faiss-split-feedstock/pull/17/commits/8e8a3e7c276a3b90bc510706ec2f776d1a6fd483) to gcc=7 on linux, or tackling https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/1000.

recipe/install_nvcc.sh
```diff
@@ -124,6 +124,11 @@ EOF
 mkdir -p "${PREFIX}/bin"
 cat > "${PREFIX}/bin/nvcc" <<EOF
 #!/bin/bash
-"\${CUDA_HOME}/bin/nvcc" -ccbin "\${CXX}" \$@
+for i ; do
+  case \$i in -ccbin)
+    exec "\${CUDA_HOME}/bin/nvcc" "\${@}"
+  esac
+done
```
jakirkham, line 131: Would it make sense to just loop through `${@}` and see if `-ccbin` is there without the `exec`? Might simplify things a bit 🙂

jakirkham (commented):
Thanks for working on this Marcell! 😄

Generally this makes sense. CMake has gotten a bit smarter with setting these things. So it needs less hand-holding, which is nice to see 🙂

Made one suggestion below

recipe/install_nvcc.sh
mbargull: > Would it make sense to just loop through `${@}`

This is exactly what `for i` does ;)
> without the `exec`?

The `exec` is just an independent improvement: If we have a launcher script, it makes sense to `exec` the called process so that our shell process can be stopped/replaced.

recipe/meta.yaml
```diff
@@ -24,17 +24,24 @@ outputs:
     requirements:
       host:
         # Needed to symlink libcuda into sysroot libs.
-        - {{ compiler("c") }}
+        # FIXME: nvcc 9.* and 10.* don't work with GCC 9.
+        #        Adding custom GCC versions via conda_build_config.yaml fails
+        #        due to https://github.com/conda/conda-build/issues/3974
+        #        => use hardcoded compiler package names to workaround that.
+        - {{ compiler("c") }}    # [not (linux64 and (cuda_compiler_version or "").startswith(("9.", "10.")))]
+        - gcc_linux-64 7         # [linux64 and (cuda_compiler_version or "").startswith(("9.", "10."))]
```
kkraus14, line 32: Does this have a `run_export`? If so can we do `<= 7` so users can use an older compiler if desired still?

Similarly, it would be good to handle the CUDA 11.0, and 11.1 requirements: https://github.com/conda-forge/nvcc-feedstock/issues/51

Something like:
```suggestion
        - gcc_linux_64 <=7       # [linux64 and (cuda_compiler_version or "").startswith(("9.", "10."))]
        - gcc_linux_64 <=9       # [linux64 and (cuda_compiler_version or "").startswith("11.0")]
        - gcc_linux_64 <=10     # [linux64 and (cuda_compiler_version or "").startswith("11.1")]
```

recipe/install_nvcc.sh
jakirkham: Interesting would have expected that to look more like `for i in "${@}"`. It might be clearer to write that (even if `bash` is doing some magic under-the-hood).

mbargull: > (even if `bash` is doing some magic under-the-hood).

It's standard POSIX shell: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_04_03

> Interesting would have expected that to look more like `for i in "${@}"`. It might be clearer to write that

I don't mind either way -- I can add a commit to change that :).

recipe/meta.yaml
mbargull: This is just the compiler version used during the build/test. We only have `run_exports` for the runtimes (`libgcc-ng`/`libstdcxx-ng`) but not for the compilers themselves. You should be able to build downstream packages with older compilers either way.
I've used the exact versions because that's how we build packages on conda-forge generally, with a fixed compiler version (`7` or `9` recently). I've also not special cased `nvcc=10.1|10.2` to have a consistent version with other packages (we hadn't pinned GCC `8` before).

My goal was to do the least necessary for the build and rather wait until someone fixes the mentioned conda-build bug instead of investing time in workarounds.
That said, I think it's best if I split this change off into another PR and just use a temporary workaround in the CI files for the intended change of this PR.

mbargull: I hadn't looked at the whole recipe until now...
I don't really understand the purpose of the `{{ compiler('c') }}` dependency in `host` here. It is not used in any way.
Also the other duplicate deps for sysroot/glibc I dont understand either.

I can't deduce what the intentions of the dependency declarations overall are. They need some cleanup, but without knowing the indented use, I can't help with that.

mbargull:
> If needed, I'll handle the Windows part once this PR is merged!

Thanks @jaimergp, much appreciated!

mbargull:
@conda-forge/nvcc, this is ready for review.

I opted to postpone the unrelated GCC version pinning to a later PR (<-- not opened yet, because I had problems understanding what the desired actions would be, see https://github.com/conda-forge/nvcc-feedstock/pull/54#discussion_r537824637 ).

recipe/install_nvcc.sh
jakirkham: It would make it clearer I think. We already are `cat`ing into a script (my fault 😞). So it's already a bit confusing. The less confusing we can make it the better 🙂

recipe/install_nvcc.sh
```diff
@@ -122,8 +122,14 @@ EOF
...
+cat > "${PREFIX}/bin/nvcc" <<'EOF'
 #!/bin/bash
-"\${CUDA_HOME}/bin/nvcc" -ccbin "\${CXX}" \$@
+for arg in "${@}" ; do
+  case ${arg} in -ccbin)
+    # If -ccbin argument is already provided, don't add an additional one.
+    exec "${CUDA_HOME}/bin/nvcc" "${@}"
+  esac
+done
+exec "${CUDA_HOME}/bin/nvcc" -ccbin "${CXX}" "${@}"
```
kkraus14, line 133: If we find `-ccbin` in the arguments, will this execute `nvcc` twice? The first time within thise loop, and then again after the loop?

recipe/meta.yaml
kkraus14: `nvcc` uses the host compiler during it's compilation. Depending on the CUDA version, there's compatibilities with the version of the host compiler used that we need to respect.

Additionally, as of CUDA 11, CentOS 6 is no longer supported so we needed a way to force only CentOS 7+.

recipe/install_nvcc.sh
mbargull: `exec` replaces the current process (i.e., the wrapper script) with the actual `nvcc`.

recipe/meta.yaml
mbargull: > `nvcc` uses the host compiler during it's compilation. Depending on the CUDA version, there's compatibilities with the version of the host compiler used that we need to respect.

This recipe only creates wrapper/(de-)activation scripts => the compiler is not used. If it's meant to be declared for downstream packages, something has to be declared in `requirements/run`.

> Additionally, as of CUDA 11, CentOS 6 is no longer supported so we needed a way to force only CentOS 7+.

The duplication of that in `host`/`run`/`test/requires` is what I meant.

recipe/install_nvcc.sh
jakirkham: Yeah it shouldn't. Though agree it might be clearer if we had a simple `if...else` here.

mbargull: To me all of this looks very clear -- but well, I've just messed about with POSIX-y shell code a lot the last few years with all these packaging stuff...
If there are code comments or changes you'd like to have which would help comprehension for the maintainers, i.e., you, please add
````
```suggestion
```
````
blocks I can apply.
(If you're unsure about shell syntax, then I can also just translate pseudo code on the fly if you give me some.)

kkraus14: TIL about `exec`. This is fine @mbargull, resolving this.

kkraus14 (approved):

kkraus14:
Will let @jakirkham take another look as well before merging

kkraus14:
Merging as no response from others.

PR #47, opened 2020-10-08
Use CUDA 11.0 migrator

jakirkham:
Instead of adding a `conda_build_config.yaml` to affect the manual migration of `nvcc`, simply add the same CUDA 11.0 migrator as used on other feedstocks. The end result is the same, but it allows this feedstock to match others a bit more closely and simplify things here

Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

xref: https://github.com/conda-forge/nvcc-feedstock/pull/43
xref: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/794

PR #55, opened 2020-12-08
restrict host compiler version based on CUDA version

### LINKED ISSUE #51: nvcc package should constrain c / cxx compiler versions based on cuda version

kkraus14:
Checklist

Closes #51

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -20,22 +20,27 @@ outputs:
       run_exports:
         strong:
           - cudatoolkit {{ cuda_compiler_version }}|{{ cuda_compiler_version }}.*
-          - __glibc >=2.17  # [linux and cuda_compiler_version == "11.0"]
+          - __glibc >=2.17  # [linux and (cuda_compiler_version or "").startswith("11.")]
     requirements:
       host:
         # Needed to symlink libcuda into sysroot libs.
         - {{ compiler("c") }}
-        - sysroot_linux-64 2.17  # [linux and cuda_compiler_version == "11.0"]
-        - __glibc >=2.17         # [linux and cuda_compiler_version == "11.0"]
+        - sysroot_linux-64 2.17  # [linux and (cuda_compiler_version or "").startswith("11.")]
+        - __glibc >=2.17         # [linux and (cuda_compiler_version or "").startswith("11.")]
       run:
-        - sysroot_linux-64 2.17  # [linux and cuda_compiler_version == "11.0"]
-        - __glibc >=2.17         # [linux and cuda_compiler_version == "11.0"]
+        - sysroot_linux-64 2.17  # [linux and (cuda_compiler_version or "").startswith("11.")]
+        - __glibc >=2.17         # [linux and (cuda_compiler_version or "").startswith("11.")]
+      run_constrained:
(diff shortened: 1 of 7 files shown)


kkraus14:

recipe/meta.yaml
```diff
@@ -20,22 +20,27 @@ outputs:
...
       run:
-        - sysroot_linux-64 2.17  # [linux and cuda_compiler_version == "11.0"]
-        - __glibc >=2.17         # [linux and cuda_compiler_version == "11.0"]
+        - sysroot_linux-64 2.17  # [linux and (cuda_compiler_version or "").startswith("11.")]
+        - __glibc >=2.17         # [linux and (cuda_compiler_version or "").startswith("11.")]
+      run_constrained:
+        - gcc_linux-64 <=10      # [linux and (cuda_compiler_version or "").startswith("11.1.")]
+        - gcc_linux-64 <=9       # [linux and (cuda_compiler_version or "").startswith("11.0.")]
+        - gcc_linux-64 <=8       # [linux and (cuda_compiler_version or "").startswith(("10.2.", "10.1."))]
+        - gcc_linux-64 <=7       # [linux and (cuda_compiler_version or "").startswith(("10.0.", "9.2."))]
```
kkraus14, line 37: Using run constraints felt more appropriate than run dependencies to continue to allow people to bring their own compiler

jakirkham: I don't think this will work as the version is already set in the recipe's build variants, which conflicts.

My guess is we will need to use [`zip_keys` with the `cxx_compiler_version` in `conda-forge-pinning`]( https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/53b26ee6df5eb51181b4ca74fdfb57940316be5c/recipe/conda_build_config.yaml#L131-L143 ) like what we have done already with `docker_image`. Though we can test this hypothesis out by adding a `conda_build_config.yaml` to the `recipe` here with the same change.

kkraus14: I'm a bit lost as to how I would go about doing this. Would I create a new key with say `[9, 7, 7, 7, 9]` to match the `cuda_compiler_version` keys? How would I make `{{ compiler("c") }}` play nicely with that since it seems to already be set by the earlier zip key?

h-vetinari: I didn't manage to get that particular zip working: https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/1000

kkraus14:
Closing as this approach won't work. Will open a PR in `conda-forge-pinnings`.

jakirkham:
Sounds good.

Yeah to your earlier question, we would need to combine them into the same `zip_keys` group. Shouldn't be too difficult. We can also build a migrator on the same premise and test that out here as well ( https://github.com/conda-forge/nvcc-feedstock/pull/47 ). That should help us build confidence that it is doing the right thing.

PR #58, opened 2020-12-29
Add some more env vars

jaimergp:
Checklist
* [X] Ensured the license file is being packaged.

This addresses several limitations found in build systems:

* All: CMake looks for CUDA_PATH, not CUDA_HOME, when trying to identify the CUDA installation
* CUDA 11+ does not package `version.txt`, so the version detection logic has been updated to use `nvcc --version` instead.
* Windows: MSBuild integrations might require a `CudaToolkitDir` env var to work properly
* Windows: Define INCLUDE, although CMake should know how to add that too.

Some more questions:

* Windows: Shall we get rid of CFLAGS and friends? They are causing plenty of trouble (double quotes, backslashes), and solve nothing since CMake handles those just ok... Some examples seen in https://github.com/conda-forge/faiss-split-feedstock/pull/19, https://github.com/conda-forge/magma-feedstock/pull/7, https://github.com/conda-forge/prismatic_split-feedstock/pull/18
  * Removed.
* Linux: Unconfirmed, maybe only old-style CMake. Sometimes, you need to link to the stubs instead of the real libraries, which requires `export CUDA_LIB_PATH="${CUDA_PATH}/lib64/stubs"`. Should we define that too?
  * Handled via `CMAKE_ARGS`.


jaimergp:

jaimergp:

kkraus14 (approved):

kkraus14:
@jaimergp one note here: If someone enables the `CUDA` language in a CMake project it will still incorrectly find `nvcc` first unless we explicitly pass a CMake variable of `CUDAToolkit_ROOT` (can't use environment variable as compiler search for `CUDA` language overrides it).

jaimergp:
> @jaimergp one note here: If someone enables the `CUDA` language in a CMake project then `FindCUDAToolkit` will still incorrectly find our `nvcc` shim unless we explicitly pass a CMake variable of `CUDAToolkit_ROOT` (can't use environment variable as compiler search for `CUDA` language overrides it).

Shall we add `-DCUDAToolkit_ROOT=${CUDA_PATH}` to `$CMAKE_ARGS`?

recipe/windows/activate.bat
```diff
@@ -70,13 +76,17 @@ if errorlevel 1 (
 
 set "CUDA_PATH=%CUDA_PATH%"
 set "CUDA_HOME=%CUDA_PATH%"
+set "CudaToolkitDir=%CUDA_PATH%"
 
 :: These paths might include double quotes, so we need this syntax
 :: to avoid quote poisoning
 (set CFLAGS=%CFLAGS% -I"%CUDA_HOME%\include")
 (set CPPFLAGS=%CPPFLAGS% -I"%CUDA_HOME%\include")
 (set CXXFLAGS=%CXXFLAGS% -I"%CUDA_HOME%\include")
```
isuruf, line 85: Let's delete these now that `INCLUDE` is set

kkraus14:
> Shall we add `-DCUDAToolkit_ROOT=${CUDA_PATH}` to `$CMAKE_ARGS`?

That will cause warnings for all projects which aren't using `FindCUDAToolkit` which is definitely non-ideal. I'd be more comfortable just acknowledging and documenting it for now.

isuruf:
> That will cause warnings for all projects which aren't using FindCUDAToolkit which is definitely non-ideal.

There are more of that in CMAKE_ARGS, so it's fine to have it. For eg: CMAKE_INSTALL_LIBDIR=lib is there but there'll be a warning if the project doesn't use `GNUInstallDirs` module in CMake.

jaimergp:
I'd also like to note that on CUDA 11+, `CUDA_HOME/version.txt` does not exist, which causes the activation script to exit early. I suggest something like this, instead:

```
if [[ -z "$(${CUDA_HOME}/bin/nvcc --version | grep 'Cuda compilation tools, release 11.0')" ]]
then
  echo "Version of installed CUDA didn't match package"
  return 1
fi
```

recipe/install_nvcc.sh
```diff
@@ -53,17 +63,34 @@ then
...
+
+# CMake looks up components in CUDA_PATH, not CUDA_HOME
+export CUDA_PATH="\${CUDA_HOME}"
+# New-style CUDA integrations in CMake
+CMAKE_ARGS="\${CMAKE_ARGS:-} -DCUDAToolkit_ROOT=\${CUDA_HOME}"
+# Old-style CUDA integrations in CMake
+## See https://github.com/conda-forge/nvcc-feedstock/pull/58#issuecomment-752179349
+CMAKE_ARGS+=" -DCUDA_TOOLKIT_ROOT_DIR=\${CUDA_HOME}"
+## Prevent errors like https://github.com/floydhub/dl-docker/issues/59
+CMAKE_ARGS+=" -DCMAKE_LIBRARY_PATH=\${CUDA_HOME}/lib64/stubs"
```
isuruf, line 87: This is bad. I don't think you need this

recipe/install_nvcc.sh
```diff
@@ -53,17 +63,34 @@ then
...
+## Avoid https://github.com/conda-forge/openmm-feedstock/pull/44#issuecomment-753560234
+CMAKE_ARGS+=" -DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=BOTH"
```
isuruf, line 89: Can you see if adding `$CUDA_HOME` to `CMAKE_PREFIX_PATH` env var works?

recipe/install_nvcc.sh
jaimergp: Hm, ok. I think it's no longer needed in linux 64, but I recall running into it with PowerPC. I can't recall how or why, so I'll remove it for now.

This might have the same effect and less hacky, if needed:

`export CUDA_LIB_PATH="${CUDA_PATH}/lib64/stubs"`

recipe/install_nvcc.sh
jaimergp: It does not work, at least for OpenMM.

isuruf: Can you try fixing `CMAKE_FIND_ROOT_PATH`? It's currently https://github.com/conda-forge/ctng-compiler-activation-feedstock/blob/master/recipe/activate-gcc.sh#L142

jaimergp: I am trying by doing this:

```
CMAKE_ARGS="$(echo ${CMAKE_ARGS} | sed -E -e "s|(-DCMAKE_FIND_ROOT_PATH=\S+)|\1;$CUDA_HOME|")"
```

Which seems to work, but with warnings:

```
-- Found CUDA: /usr/local/cuda (found version "11.0") 
CMake Warning (dev) at /home/conda/feedstock_root/build_artifacts/openmm_1609778103608/_build_env/share/cmake-3.19/Modules/FindPackageHandleStandardArgs.cmake:424 (message):
  The package name passed to `find_package_handle_standard_args` (OPENCL)
  does not match the name of the calling package (OpenCL).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  cmake_modules/FindOpenCL.cmake:86 (find_package_handle_standard_args)
  CMakeLists.txt:338 (FIND_PACKAGE)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found OPENCL: /home/conda/feedstock_root/build_artifacts/openmm_1609778103608/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/libOpenCL.so  
-- Configuring done
CMake Warning at plugins/cudacompiler/CMakeLists.txt:71 (ADD_LIBRARY):
  Cannot generate a safe runtime search path for target OpenMMCudaCompiler
  because files in some directories may conflict with libraries in implicit
  directories:

    runtime library [libnvrtc.so.11.0] in /home/conda/feedstock_root/build_artifacts/openmm_1609778103608/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib may be hidden by files in:
      /usr/local/cuda/lib64

  Some of these libraries may not be found correctly.
```

isuruf: Try adding `CUDA_HOME/include` instead of `CUDA_HOME`

jaimergp: That takes us back to square one:

```
-- Configuring done
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
CUDA_TOOLKIT_INCLUDE (ADVANCED)
   used as include directory in directory /home/conda/feedstock_root/build_artifacts/openmm_1609778103608/work/plugins/cudacompiler
   used as include directory in directory /home/conda/feedstock_root/build_artifacts/openmm_1609778103608/work/plugins/cudacompiler
   used as include directory in directory /home/conda/feedstock_root/build_artifacts/openmm_1609778103608/work/plugins/cudacompiler
   used as include directory in directory /home/conda/feedstock_root/build_artifacts/openmm_1609778103608/work/plugins/cudacompiler
   used as include directory in directory /home/conda/feedstock_root/build_artifacts/openmm_1609778103608/work/plugins/cudacompiler
   used as include directory in directory /home/conda/feedstock_root/build_artifacts/openmm_1609778103608/work/plugins/cudacompiler
   used as include directory in directory /home/conda/feedstock_root/build_artifacts/openmm_1609778103608/work/plugins/cudacompiler
   used as include directory in directory /home/conda/feedstock_root/build_artifacts/openmm_1609778103608/work/plugins/cudacompiler

CMake Error in plugins/cudacompiler/CMakeLists.txt:
  Found relative path while evaluating include directories of
  "OpenMMCudaCompiler":

    "CUDA_TOOLKIT_INCLUDE-NOTFOUND"



CMake Error in plugins/cudacompiler/CMakeLists.txt:
  Found relative path while evaluating include directories of
  "OpenMMCudaCompiler":

    "CUDA_TOOLKIT_INCLUDE-NOTFOUND"
```

isuruf: Can you try adding CUDA_HOME to the front?

jaimergp: That seems to work. With:

```
CMAKE_ARGS="$(echo ${CMAKE_ARGS} | sed -E -e "s|(-DCMAKE_FIND_ROOT_PATH=)(\S+)|\1$CUDA_HOME;\2|")"
```

We get:

```
-- Found CUDA: /usr/local/cuda (found version "11.0") 
CMake Warning (dev) at /home/conda/feedstock_root/build_artifacts/openmm_1609778103608/_build_env/share/cmake-3.19/Modules/FindPackageHandleStandardArgs.cmake:424 (message):
  The package name passed to `find_package_handle_standard_args` (OPENCL)
  does not match the name of the calling package (OpenCL).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  cmake_modules/FindOpenCL.cmake:86 (find_package_handle_standard_args)
  CMakeLists.txt:338 (FIND_PACKAGE)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found OPENCL: /home/conda/feedstock_root/build_artifacts/openmm_1609778103608/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/libOpenCL.so  
-- Configuring done
-- Generating done
```

jaimergp: See last commits for this addition.

kkraus14 (approved):
lgtm, will let @isuruf take another look before merging

recipe/install_nvcc.sh
```diff
@@ -53,17 +63,33 @@ then
...
+# CMake looks up components in CUDA_PATH, not CUDA_HOME
+export CUDA_PATH="\${CUDA_HOME}"
+# New-style CUDA integrations in CMake
+CMAKE_ARGS="\${CMAKE_ARGS:-} -DCUDAToolkit_ROOT=\${CUDA_HOME}"
+# Old-style CUDA integrations in CMake
+## See https://github.com/conda-forge/nvcc-feedstock/pull/58#issuecomment-752179349
+CMAKE_ARGS+=" -DCUDA_TOOLKIT_ROOT_DIR=\${CUDA_HOME}"
+## Avoid https://github.com/conda-forge/openmm-feedstock/pull/44#issuecomment-753560234
+## We need CUDA_HOME in _front_ of CMAKE_FIND_ROOT_PATH
+CMAKE_ARGS="\$(echo \${CMAKE_ARGS} | sed -E -e "s|(-DCMAKE_FIND_ROOT_PATH=)(\S+)|\1\$CUDA_HOME;\2|")"
```
isuruf, line 88: Can you add sed to `run` requirements?

recipe/install_nvcc.sh
```diff
@@ -53,17 +63,33 @@ then
...
 fi
 
-if [[ \$(grep -q "CUDA Version ${PKG_VERSION}" \${CUDA_HOME}/version.txt) -ne 0 ]]
+if [[ -z "\$(\${CUDA_HOME}/bin/nvcc --version | grep "Cuda compilation tools, release ${PKG_VERSION}")" ]]
 then
-    echo "Version of installed CUDA didn't match package"
-    return 1
+  echo "Version of installed CUDA didn't match package"
+  return 1
 fi
```
jakirkham, line 70: Does this need to change?

recipe/install_nvcc.sh
jakirkham: This appears to be done below. Marking as resolved.

jakirkham (commented):
Thanks Jaime! 😄

Had a question about one of the changes below.

recipe/install_nvcc.sh
kkraus14: There was a comment earlier that `version.txt` doesn't exist properly in all CUDA installations

jakirkham: Ah ok. Sorry for missing that then

jakirkham (approved):

jakirkham:
Thanks Jaime! And everyone for the reviews! 😄

PR #59, opened 2021-01-08
Add CMake tests (and potentially more env vars)

jaimergp:
Checklist

After some more errors observed downstream (even with last PR merged, see https://github.com/conda-forge/arrow-cpp-feedstock/pull/247), I have decided we should test against some simple CMake projects to make sure our configuration works instead of guessing. This PR provides those and will make them work.

Note how:

* Modern CMake still requires some more configuration (or maybe my CMakeLists is misconfigured?)
* Windows might need all forward slashes

### DIFF
--- a/recipe/cmake-tests/CMakeLists.txt
+++ b/recipe/cmake-tests/CMakeLists.txt
@@ -1,0 +1,25 @@
+cmake_minimum_required(VERSION 3.8)
+
+project(diana)
+
+option(MODERN_CUDA "Find CUDA with modern approach" ON)
+
+file(GLOB cpu_source_files "${CMAKE_SOURCE_DIR}/src/*.cc")
+file(GLOB gpu_source_files "${CMAKE_SOURCE_DIR}/src/*.cu")
+
+
+if(MODERN_CUDA)
+    # Modern CMake
+    set(CMAKE_CUDA_STANDARD 17)
+    set(CMAKE_CUDA_STANDARD_REQUIRED TRUE)
+    enable_language(CUDA)
+    add_executable(diana ${cpu_source_files} ${gpu_source_files})
+else()
+    # Old CMake
+    add_executable(diana ${cpu_source_files})
+    find_package(CUDA REQUIRED)
+    set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -std=c++17")
+    CUDA_ADD_LIBRARY(diana_gpu ${gpu_source_files})
+    target_link_libraries(diana diana_gpu)
+endif()
+
--- a/recipe/test_nvcc.sh
+++ b/recipe/test_nvcc.sh
@@ -64,8 +64,34 @@ if [[ "${CFLAGS}" == "${CFLAGS_CONDA_NVCC_TEST}" ]]; then
 else
     echo "CFLAGS is incorrectly set to '${CFLAGS}', should be set to '${CFLAGS_CONDA_NVCC_TEST}'" && exit 1
 fi
+
 # Reactivate
 source ${PREFIX}/etc/conda/activate.d/${PKG_NAME}_activate.sh
 
 # Try building something
 nvcc test.cu
+
+# Try different CMake setups
+cd cmake-tests/
+
+# Old-style CMake (deprecated FindCUDA)
+set -x
+rm -rf build || true
+mkdir -p build
+cd build
+cmake ${CMAKE_ARGS} -DMODERN_CUDA=OFF ..
+make
+./diana
+cd ..
+rm -rf build
+
+# New-style CMake (FindCUDAToolkit)
+mkdir -p build
+cd build
+export CUDACXX=/usr/local/cuda/bin/nvcc
+export CUDAHOSTCXX=${CC}
+cmake ${CMAKE_ARGS} -DMODERN_CUDA=ON .. \
+    # -DCMAKE_CUDA_COMPILER=/usr/local/cuda/bin/nvcc \
+    # -DCMAKE_CUDA_HOST_COMPILER=${CXX} \
+make
+./diana
--- a/recipe/windows/test_nvcc.bat
+++ b/recipe/windows/test_nvcc.bat
@@ -71,3 +71,28 @@ if errorlevel 1 exit 1
 
 :: Try building something
 nvcc test.cu
+
+
+:: Try different CMake setups
+cd cmake-tests/
+
+:: Old-style CMake (deprecated FindCUDA)
+mkdir build
+cd build
+cmake %CMAKE_ARGS% -DMODERN_CUDA=OFF ..
+jom
+./diana
+cd ..
+rmdir /s /q build
+
+:: New-style CMake (FindCUDAToolkit)
+mkdir build
+cd build
+set "CUDACXX=%CUDA_PATH%\bin\nvcc.exe"
+set "CUDAHOSTCXX=%CC%"
+cmake -DMODERN_CUDA=ON ..
+:: -DCMAKE_CUDA_COMPILER="%CUDA_PATH%\bin\nvcc.exe" ^
+:: -DCMAKE_CUDA_HOST_COMPILER="%CC%" ^
+
+jom
+./diana
--- a/recipe/cmake-tests/LICENSE
+++ b/recipe/cmake-tests/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2017 Luke Yeager
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/recipe/cmake-tests/README.md
+++ b/recipe/cmake-tests/README.md
@@ -1,0 +1,1 @@
+Tests adapted from https://github.com/lukeyeager/cmake-cuda-example
--- a/recipe/cmake-tests/src/gpu.cu
+++ b/recipe/cmake-tests/src/gpu.cu
@@ -1,0 +1,15 @@
+#include <iostream>
+#include "gpu.hpp"
+
+void printCudaVersion()
+{
+    std::cout << "CUDA Compiled version: " << __CUDACC_VER_MAJOR__ << "." << __CUDACC_VER_MINOR__ << std::endl;
+
+    int runtime_ver;
+    cudaRuntimeGetVersion(&runtime_ver);
+    std::cout << "CUDA Runtime version: " << runtime_ver << std::endl;
+
+    int driver_ver;
+    cudaDriverGetVersion(&driver_ver);
+    std::cout << "CUDA Driver version: " << driver_ver << std::endl;
+}
--- a/recipe/cmake-tests/src/gpu.hpp
+++ b/recipe/cmake-tests/src/gpu.hpp
@@ -1,0 +1,1 @@
+void printCudaVersion();
--- a/recipe/cmake-tests/src/main.cc
+++ b/recipe/cmake-tests/src/main.cc
@@ -1,0 +1,10 @@
+#include <iostream>
+#include "gpu.hpp"
+
+int main()
+{
+    std::cout << "Hello, CUDA!" << std::endl;
+    printCudaVersion();
+
+    return 0;
+}
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "nvcc" %}
-{% set number = 11 %}
+{% set number = 12 %}
 
 package:
   name: "{{ name }}"
@@ -38,9 +38,12 @@ outputs:
         - {{ compiler("c") }}
         # Host code is forwarded to a C++ compiler
         - {{ compiler("cxx") }}
+        - cmake
+        - jom  # [win]
         - sysroot_linux-64 2.17  # [linux and cuda_compiler_version == "11.0"]
       files:
         - test.cu
+        - cmake-tests
         - test_nvcc.sh  # [linux]
         - windows\test_nvcc.bat  # [win]
       commands:


recipe/cmake-tests/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+    enable_language(CUDA)
+    add_executable(diana ${cpu_source_files} ${gpu_source_files})
+else()
+    # Old CMake
+    add_executable(diana ${cpu_source_files})
+    find_package(CUDA REQUIRED)
+    set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -std=c++17")
+    CUDA_ADD_LIBRARY(diana_gpu ${gpu_source_files})
+    target_link_libraries(diana diana_gpu)
+endif()
```
kkraus14, line 24: Could we also test the `FindCUDAToolkit` approach? I.E. `find_package(CUDAToolkit)`

Ideally we should test that `FindCUDA` and `FindCUDAToolkit` both work with and without doing `enable_language(CUDA)` which changes the behavior.

jaimergp: Hm, so the full matrix would be:

* FindCUDA
* FindCUDA + enable_language
* FindCUDAToolkit
* FindCUDAToolkit + enable_language

right?

kkraus14: I would add just `enable_language` as well.

jaimergp: _FindCUDAToolkit without enable_language_  does not seem to be a valid combination.

kkraus14: > _FindCUDAToolkit without enable_language_ does not seem to be a valid combination.

It absolutely should be and I have projects that do exactly this. It sounds like we're not setting a necessary `env` var or we're doing something that causes `FindCUDAToolkit` to fail.

jaimergp:
Ok, @kkraus14, I added the needed bits to test all four cases, but FindCUDAToolkit + enable_language fails.

I can't use `cuda_add_library` since that's only available with `FindCUDA`, so I am using plain `add_library`. WIthout `enable_language(CUDA)` it fails due to a missing reference. Inspecting the built libraries, well, it just doesn't build. I guess my CMake config is wrong and since I barely know what I am doing, I'll stop trying random combinations... I hope the code is clear for somebody else to take over though :)

One more thing to note: in some successful builds, the CUDA runtime version is reported as zero. This only happens when enable_language is OFF. This does not throw an error, but I guess that's a wrong result...

recipe/cmake-tests/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+    add_executable(diana ${cpu_source_files})
+
+    if("${CUDA_FINDER}" STREQUAL "CUDA")
+        message(STATUS "With CUDA")
+        find_package("${CUDA_FINDER}" REQUIRED)
+        set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -Xcompiler ,\"-std=c++14\"")
+        CUDA_ADD_LIBRARY(diana_gpu ${gpu_source_files})
+    else("${CUDA_FINDER}" STREQUAL "CUDA")
+        find_package("${CUDA_FINDER}" REQUIRED)
+        add_library(diana_gpu ${gpu_source_files})
```
kkraus14, line 38: If `enable_language(CUDA)` isn't called it's not valid to try to build `.cu` files. It would be valid to build CPU source files `.cc` that use CUDA headers / libraries, but we'd need to add a `target_link_libraries` for example.

Alternatively, I think the main thing we want to test here is that the targets / variables created by `FindCUDAToolkit` / `FindCUDA` are populated with values as expected, so we don't necessarily need to compile anything.

kkraus14: Instead of doing `add_library` here I think it would be good enough to check that `CUDAToolkit_LIBRARY_ROOT` (requires CMake 3.18+) is set as expected for `FindCUDAToolkit`.

SImilarly, we be good to check that `CUDA_TOOLKIT_ROOT_DIR` is set as expected for `FindCUDA` above in addition to doing the `CUDA_ADD_LIBRARY`.

recipe/windows/test_nvcc.bat
```diff
@@ -71,3 +71,22 @@ if errorlevel 1 exit 1
...
 nvcc test.cu
+
+
+:: Try different CMake setups
+cd cmake-tests/
+
+for %%F in (
+    "-DWITH_FINDCUDA=ON -DWITH_FINDCUDATOOLKIT=OFF -DWITH_ENABLE_LANGUAGE=OFF"
+    "-DWITH_FINDCUDA=ON -DWITH_FINDCUDATOOLKIT=OFF -DWITH_ENABLE_LANGUAGE=ON"
+    "-DWITH_FINDCUDA=OFF -DWITH_FINDCUDATOOLKIT=ON -DWITH_ENABLE_LANGUAGE=ON"
```
kkraus14, line 82: I think these are the 5 cases we want to test, and in each case we want to make sure that CUDA related paths are being set as expected, but that CMake is using our `nvcc` wrapper instead of the `nvcc` binary directly:
```suggestion
    "-DWITH_FINDCUDA=OFF -DWITH_FINDCUDATOOLKIT=OFF -DWITH_ENABLE_LANGUAGE=ON"
    "-DWITH_FINDCUDA=ON -DWITH_FINDCUDATOOLKIT=OFF -DWITH_ENABLE_LANGUAGE=OFF"
    "-DWITH_FINDCUDA=OFF -DWITH_FINDCUDATOOLKIT=ON -DWITH_ENABLE_LANGUAGE=OFF"
    "-DWITH_FINDCUDA=ON -DWITH_FINDCUDATOOLKIT=OFF -DWITH_ENABLE_LANGUAGE=ON"
    "-DWITH_FINDCUDA=OFF -DWITH_FINDCUDATOOLKIT=ON -DWITH_ENABLE_LANGUAGE=ON"
```

jaimergp: Windows tests were not running because the nvcc wrapper returns right away and exits the script. Apparently this needs to be run with `call nvcc ...`. My love for cmd won't stop growing! 

I have amended the Windows tests now and added different generators too.

recipe/cmake-tests/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+    if("${CUDA_FINDER}" STREQUAL "CUDA")
+        find_package("${CUDA_FINDER}" REQUIRED)
+        set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -Xcompiler ,\"-std=c++14\"")
+        add_definitions("-DWITHCUFILES")
+        CUDA_ADD_LIBRARY(diana_gpu ${gpu_source_files})
+        target_link_libraries(diana diana_gpu)
+    elseif("${CUDA_FINDER}" STREQUAL "CUDAToolkit")
+        # This example does not link to .cu libraries
+        find_package("${CUDA_FINDER}" REQUIRED)
+        target_link_libraries(diana CUDA::cudart)
```
kkraus14, line 43: Could we check the result of `CUDAToolkit_NVCC_EXECUTABLE` (https://cmake.org/cmake/help/git-stage/module/FindCUDAToolkit.html#result-variables) to ensure it's pointing to the conda `nvcc` as opposed to the system install? I've seen cmake projects that integrate non-cmake libraries have to manually invoke `nvcc` in order to build things and we'd want them to similarly use the conda wrapper.

recipe/cmake-tests/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+
+if(WITH_ENABLE_LANGUAGE)
+    enable_language(CUDA)
+    set(CMAKE_CUDA_STANDARD 14)
+    set(CMAKE_CUDA_STANDARD_REQUIRED TRUE)
+    if(CUDA_FINDER)
+        find_package("${CUDA_FINDER}" REQUIRED)
+    endif(CUDA_FINDER)
+    add_definitions("-DWITHCUFILES")
+    add_executable(diana ${cpu_source_files} ${gpu_source_files})
```
kkraus14, line 29: Could we check `CMAKE_CUDA_COMPILER` or something similar to ensure things are actually being built with the conda `nvcc` wrapper instead of silently using the system `nvcc`?

recipe/cmake-tests/CMakeLists.txt
kkraus14: We should check this in both cases of `enable_language(CUDA)` as if the `CUDA` language is enabled, it changes the search path that CMake uses for this module.

recipe/cmake-tests/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+
+else(WITH_ENABLE_LANGUAGE)
+    add_executable(diana ${cpu_source_files})
+
+    if("${CUDA_FINDER}" STREQUAL "CUDA")
+        find_package("${CUDA_FINDER}" REQUIRED)
+        set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -Xcompiler ,\"-std=c++14\"")
+        add_definitions("-DWITHCUFILES")
+        CUDA_ADD_LIBRARY(diana_gpu ${gpu_source_files})
```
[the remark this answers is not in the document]
kkraus14: Similarly, should test this for both `enable_language(CUDA)` cases.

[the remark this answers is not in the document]
jaimergp: Not nitpicking at all, I had to manually pass the system nvcc location to make some tests pass, so if anything I am guaranteeing that it does _not_ use our wrapper (but to the same effect, because we are passing the value of `ccbin` in another Dflag). Next round of changes will involve undoing this whenever possible...

jaimergp:
@kkraus14, I added the checks you wanted, which are failing because [I am exporting these variables](https://github.com/jaimergp/nvcc-feedstock/blob/cmake-tests/recipe/test_nvcc.sh#L95-L96) to have CMake find the CUDA root successfully. Otherwise it finds our nvcc first and then it thinks the CUDA_PATH is PREFIX. Passing several different environment variables did not help in my initial tests, but I will revisit it again.

Also, what CMake variable should I look up for the case then we don't use FindCUDA nor FindCUDAToolkit?

kkraus14:
> Also, what CMake variable should I look up for the case then we don't use FindCUDA nor FindCUDAToolkit?

`CMAKE_CUDA_COMPILER` most likely is the only thing we can check in that situation and we'd want it to point to our wrapper.

> I added the checks you wanted, which are failing because I am exporting these variables to have CMake find the CUDA root successfully. Otherwise it finds our nvcc first and then it thinks the CUDA_PATH is PREFIX. Passing several different environment variables did not help in my initial tests, but I will revisit it again.

This is exactly the problem we'd want to solve. We want to make sure that CMake uses our wrapper as the compiler command but still knows where the actual CUDA root is.

isuruf:
> This is exactly the problem we'd want to solve. We want to make sure that CMake uses our wrapper as the compiler command but still knows where the actual CUDA root is.

Do we really need this? We can just let CMae use the original nvcc and set `CMAKE_CUDA_HOST_COMPILER` to gcc which amounts to the same thing as using the nvcc wrapper.

kkraus14:
> Do we really need this? We can just let CMae use the original nvcc and set `CMAKE_CUDA_HOST_COMPILER` to gcc which amounts to the same thing as using the nvcc wrapper.

I've seen projects out there that integrate non-CMake projects that expect to have `nvcc` called directly by using `CMAKE_CUDA_COMPILER` to do so. It would be nice to support. Additionally, the problem of finding our wrapper and getting the wrong CUDA root due to it should be fixed in the next CMake release: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5620

jaimergp:
Cool! I guess we should wait until then instead of trying different (potentially hackish) workarounds here? Should `cmake` be added to `run_constrained` for that next version in that case?

jakirkham:
Looks like that is merged already and scheduled for 3.20.0. Suppose we could try using that PR to patch our own build of CMake. Not sure how doable that would be.

kkraus14:
> Cool! I guess we should wait until then instead of trying different (potentially hackish) workarounds here? Should `cmake` be added to `run_constrained` for that next version in that case?

Possibly not a problem for conda-forge, but I've typically seen projects restrict their CMake version somewhat aggressively as CMake does introduce non-trivial changes that can cause breakages with their releases. I for one would push for us to solve the problems we're currently seeing related to CMake choosing a CUDA root based on finding our `nvcc` shim.

recipe/cmake-tests/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+# Some more checks
+if(WIN32)
+    set(NVCC_WRAPPER "$ENV{PREFIX}/Library/bin/nvcc.bat")
+else(WIN32)
+    set(NVCC_WRAPPER "$ENV{PREFIX}/bin/nvcc")
+endif(WIN32)
+
+if(CUDA_FINDER)
+    if("${CUDA_FINDER}" STREQUAL "CUDA")
+        set(FOUND_NVCC "${CMAKE_CUDA_COMPILER}")
```
kkraus14, line 59: I don't think this is valid. We'd want to test for `CMAKE_CUDA_COMPILER` in the case when `enable_language(CUDA)` is used, which is not every case of `find_package(CUDA)`. We should check the undocumented `CUDA_NVCC_EXECUTABLE` variable that is set here: https://gitlab.kitware.com/cmake/cmake/-/blob/v3.19.3/Modules/FindCUDA.cmake#L766-781

If we're not overriding it with our shim using an environment variable than we potentially should.
```suggestion
        set(FOUND_NVCC "${CUDA_NVCC_EXECUTABLE}")
```

recipe/test_nvcc.sh
```diff
@@ -64,8 +64,49 @@ if [[ "${CFLAGS}" == "${CFLAGS_CONDA_NVCC_TEST}" ]]; then
...
+    ; do
+        ((count+=1))
+        echo -e "\n\n--------"
+        echo -e "Test #$count:" $flags
+        echo -e "--------\n\n"
+        rm -rf build || true
+        mkdir -p build
+        cd build
+        export CUDACXX=${CUDA_PATH}/bin/nvcc
+        export CUDAHOSTCXX=${CC}
```
kkraus14, line 96: We absolutely shouldn't need to set these, otherwise every downstream project would need to potentially set these as well.

recipe/cmake-tests/CMakeLists.txt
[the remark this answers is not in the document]
kkraus14: Resolving as we're testing things below now.

recipe/cmake-tests/CMakeLists.txt
kkraus14: Checking things below.

recipe/cmake-tests/CMakeLists.txt
kkraus14: Checking below.

recipe/cmake-tests/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
+cmake_minimum_required(VERSION 3.17)
```
kkraus14, line 1: Would suggest bumping this to 3.18 as some of the buggy behavior we're adding workarounds for didn't exist until 3.18.

recipe/test_nvcc.sh
jaimergp: This is what I needed to pass to make CMake build the project. We need to decide whether we move this to the activation script or not.

recipe/cmake-tests/CMakeLists.txt
jaimergp: Added in last commit.

recipe/test_nvcc.sh
jaimergp: I've moved it for now, I don't think it'll hurt having it there, but happy to change it if needed!

jaimergp:
Ok I applied some of the suggestions. I don't know how we can pass the tests. I'll sum up in bullet points:

* We are exporting a plethora of CMake variables to point to the correct CUDA location.
* `CUDAHOSTCXX` also points to `CC` just in case. This has the same effect as the `-ccbin` command-line argument.
* However, CMake finds our nvcc wrapper first and thus infers that the CUDA location is PREFIX. This is a bug that will be fixed in CMake 3.20. Note that it IGNORES our CMake variables altogether!

Possible solutions:

* A) If we set `CUDACXX` to point to the real nvcc, then CMake finds the right CUDA root, but it will also ignore our NVCC wrapper. Note that:
  * Although this bypasses the wrapper, it still works because CUDAHOSTCXX has the same effect as `-ccbin`. 
  * Our wrapper is still first in PATH, so non CMake systems can still do `nvcc ...` as usual.
* B) Drop the nvcc wrapper but define `CUDA_ARGS=-ccbin ....` and instruct users to include `CUDA_ARGS` in their invocations as we do for CMake and CMAKE_ARGS.

I think solution A works as we want, but we need to adapt the CMake tests to our expectations:

* FOUND_NVCC is the real CUDA
* FOUND_CUDA_ROOT is the system installation
* FOUND_CUDA_HOST_COMPILER is the same as our PREFIX-based CC.

Let me know if I am missing something!

h-vetinari:
> Possible solutions:

Cmake 3.20, has an internal [deadline](https://gitlab.kitware.com/cmake/cmake/-/milestones/76) of Feb 1st. Not sure what their average slippage is, but for 3.19 they had 5 weeks between rc1 and final, and so far, I see no rc for 3.20. In any case, it's possible that "waiting" could be an acceptable approach as well.

kkraus14:
> In any case, it's possible that "waiting" could be an acceptable approach as well.

I don't think it's safe to assume every project can upgrade to the latest CMake version cleanly so us forcing it would be somewhat problematic.

recipe/test_nvcc.sh
```diff
@@ -64,8 +64,48 @@ if [[ "${CFLAGS}" == "${CFLAGS_CONDA_NVCC_TEST}" ]]; then
...
+    "-DWITH_ENABLE_LANGUAGE=ON  -DCUDA_FINDER=OFF        " \
+    ; do
+        ((count+=1))
+        echo -e "\n\n--------"
+        echo -e "Test #$count:" $flags
+        echo -e "--------\n\n"
+        rm -rf build || true
+        mkdir -p build
+        cd build
+        export CUDACXX=${CUDA_PATH}/bin/nvcc
```
kkraus14, line 95: Hmm, I guess this is the catch-22? Without this it will find our wrapper and infer incorrect directories in the case of `enable_language(CUDA)` being used? I'll poke around and see if there's any way to get it to use our wrapper while still pointing in the correct place.

Alternatively, we can pass desired flags directly in `CUDACXX`: https://cmake.org/cmake/help/latest/envvar/CUDACXX.html

recipe/cmake-tests/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
...
+    # Test CMake is using NVCC through our wrapper
+    if("${FOUND_NVCC}" STREQUAL "${NVCC_WRAPPER}")
+        message(STATUS "NVCC found at `${NVCC_WRAPPER}`")
+    else("${FOUND_NVCC}" STREQUAL "${NVCC_WRAPPER}")
+        message(FATAL_ERROR "NVCC found (${FOUND_NVCC}) is NOT `${NVCC_WRAPPER}!`")
+    endif("${FOUND_NVCC}" STREQUAL "${NVCC_WRAPPER}")
+
+    # However, CUDA root should be the system location, not our PREFIX
+    if("${FOUND_CUDA_ROOT}" STREQUAL "$ENV{CUDA_PATH}")
+        message(STATUS "CUDA root found at `$ENV{CUDA_PATH}`")
```
kkraus14, line 76:
```suggestion
        message(STATUS "CUDA root found at `${FOUND_CUDA_ROOT}`")
```

recipe/test_nvcc.sh
jaimergp: Yep, I think so. I can push a commit without it so you can see what happens.

> Alternatively, we can pass desired flags directly in CUDACXX

This can work nicely on Linux, but quotes within variables on Windows is such a PITA and has biten us before.

kkraus14: I don't think it will work for Linux either unfortunately. Seems like CMake parses it and rips out the `--ccbin=...` if you pass it so if someone invokes `CMAKE_CUDA_COMPILER` it won't have the option

jaimergp: I have commented out the CUDACXX definitions on 59f1cf3 if you want take a look.

jakirkham:
cc @robertmaynard

jaimergp:
@kkraus14 I am working on this again now.

Tests fail with:

```
  Compiler:
  $TEST_PREFIX/bin/nvcc


  Build flags:

  Id flags:
  --keep;--keep-dir;tmp;-ccbin=$TEST_PREFIX/bin/x86_64-conda-linux-gnu-cc
  -v

  

  The output was:

  1

  nvcc fatal : redefinition of argument 'compiler-bindir'
```

I guess we cannot pass `-ccbin` anymore?

jaimergp:
Re-enabling CUDAXX makes CMake find the real NVCC (as expected), which fails our current tests. Note we are now using CMake 3.21, but nothing has really changed. We are still in the situation described by https://github.com/conda-forge/nvcc-feedstock/pull/59#issuecomment-765286239, and I still think solution A would work.

jakirkham:
Would it be possible to generate a fresh issue summarizing the problems here (either on the feedstock or perhaps on [the CMake issue tracker]( https://gitlab.kitware.com/cmake/cmake/-/issues ))? We can then ask the CMake team for help (they are interested in making the CMake + CUDA experience better :)

jaimergp:
I'll try summarize my understanding here and then we can post an issue for CMake once we agree on the writing!

Some facts about CMake and CUDA:

* CMake has dealt with CUDA stuff in different ways throughout the releases. The chronology is, more or less:
  * FindCUDA
  * FindCUDAToolkit
  * Native language support for CUDA
* For some reason, the three approaches can be combined in different ways. This is what the tests here are trying to assess.

Some facts about our nvcc wrapper:

* The nvcc wrapper finds the actual `nvcc` in the system and exposes it to `conda-build` in a friendly way. 
* It deals with some environment variables and creates an executable in the `$CONDA_PREFIX/bin` that basically forwards the calls to the real nvcc with an added parameter: `-ccbin`, which points to the conda-installed compilers.

Interactions between CMake and our wrapper:

* CMake auto-detection mechanisms can be confused by the presence of a CUDA-foreign `nvcc` executable (our wrapper) in PATH. It will think that `$CONDA_PREFIX` is `$CUDA_HOME`, but that's not true.
* There are other environment variables we can set to tell CMake where the stuff is (the real nvcc, and our compilers), without using the wrapper at all.

I don't think there's an issue with CMake now. We can set those env vars and still ship the nvcc wrapper for non-CMake builds ([solution A here](https://github.com/conda-forge/nvcc-feedstock/pull/59#issuecomment-765286239)). However, I am not sure this is also applicable for old CMake versions. I need to look whether those env vars were always supported under the different modes.

robertmaynard:
Chronological order is `FindCUDA`, Native language support for CUDA, `FindCUDAToolkit`. 

This sounds to me like an issue with the wrapper. It needs to parse the input looking for `--ccbin` do one of the following:
- Prefer the users `ccbin` value when it differs from the wrappers
- Prefer the wrappers value when it differs from the user ( maybe with a diagnostic message )

In both cases the wrapper needs to remove multiple ccbin entries and only propagate a single value to the real `nvcc`.

jaimergp:
Thanks for the correction @robertmaynard!

I'll try something with the wrapper... I can already feel the nightmares for Windows :D

jaimergp:
Hm, the current [implementation](https://github.com/jaimergp/nvcc-feedstock/blob/cmake-tests/recipe/install_nvcc.sh#L176-L182) does not add `-ccbin` if one was already provided. I wonder why we were seeing that error then...

robertmaynard:
> Hm, the current [implementation](https://github.com/jaimergp/nvcc-feedstock/blob/cmake-tests/recipe/install_nvcc.sh#L176-L182) does not add `-ccbin` if one was already provided. I wonder why we were seeing that error then...

Two things:
- You need to capture `--compiler-bindir` 
- You are only capturing `wrapper -ccbin <path>` and not `wrapper -ccbin=<path>` which is causing the failure to see the dual ccbin flags.

jaimergp:
Ohh, good catch. Time to brush up that regex then.

jaimergp:
The duplicated ccbin problem has not appeared again, but we still see one error: the found NVCC is not our wrapper. This happens regardless CUDACXX is set or not, and regardless the value (real NVCC vs wrapper) :/

jakirkham:
cc @robertmaynard

jaimergp:
Looks like CUDAHOSTCXX has no effect if the CUDA finder is FindCUDAToolkit. I'll try with the CMake flag directly too...

jaimergp:
Ok, some updates.

Given the chronology stated some comments above, I guess that `FindCUDAToolkit` makes no sense without setting CUDA as a language. If that's correct (and the CI seems to agree), the sensible tests to check are:

* Did CMake find CUDA_HOME? Does it match our env var? (/usr/local/cuda)
* Did CMake find NVCC? We would like this to be our wrapper, but right now it will find the real NVCC thanks to CUDACXX var.
* Did CMake configure CXX as the NVCC host compiler? This is what our wrapper hardcodes anyway.

If these are the tests, then we are passing almost all of them. The NVCC host compiler will NOT be defined if FindCUDAToolkit is used without enabling CUDA as a language too. This might be by design? If a project calls NVCC directly might have trouble getting the NVCC host compiler configured correctly because it is bypassing the wrapper right now...

***

Interestingly, if the CUDA language is not enabled, CMake will find another CUDA root (`$CUDA_HOME/include`), but only in CUDA 11+? I don't know why at this point. See how it [works with CUDA 9.2](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=355566&view=logs&j=9a864fd9-6c8f-52ca-79ce-2aa6dca1a1de&t=10fc5aa2-324e-5982-4c88-6b31fcab16b3&l=585) (even if `*/include` was found first) but [fails in 11](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=355566&view=logs&j=bb1c2637-64c6-57bd-9ea6-93823b2df951&t=350df31b-3291-5209-0bb7-031395f0baa1&l=588).

robertmaynard:
> Looks like CUDAHOSTCXX has no effect if the CUDA finder is FindCUDAToolkit. I'll try with the CMake flag directly too...
> Given the chronology stated some comments above, I guess that FindCUDAToolkit makes no sense without setting CUDA as a language. If that's correct (and the CI seems to agree), the sensible tests to check are:

We need to be clear here. There is a difference between CUDA the programming language and the libraries that the CUDA Toolkit provides ( cuBLAS, cuFFT, cuRAND, ... ). These libraries can be used from `C`, `C++`, `CUDA` as they provide either a C or C++ interface. 

So it is perfectly reasonable for a C++ only project in CMake to use `FindCUDAToolkit.

Because `FindCUDAToolkit` is meant to only find the libraries it isn't directly aware of variables such as `CUDAHOSTCXX` or `CUDACXX`. Instead if the `CUDA` programming language is enabled it will try and find the associated libraries that go with that version of the CUDA toolkit. If the `CUDA` programming language isn't enabled it uses the `CUDAToolkit_ROOT` variable or the `
`CUDA_PATH`` environment variable. 


> The NVCC host compiler will NOT be defined if FindCUDAToolkit is used without enabling CUDA as a language too. This might be by design?

That is by designed as the nvcc host compiler only makes sense when the CUDA language is enabled.

jakirkham:
It's also worth noting that this package precedes the inclusion of compilers, headers, etc. as done in the newer Conda packages ( https://github.com/conda-forge/cudatoolkit-feedstock/issues/62 ) that we would like to use going forward. Am wondering how much effort we should spend towards getting CMake to combine these different pieces together vs. getting the newer Conda packages finalized and into production. WDYT?

robertmaynard:
>  Am wondering how much effort we should spend towards getting CMake to combine these different pieces together vs. getting the newer Conda packages finalized and into production. 

I would think that the new conda packages are the way forward. Those would provide a consistent interface and CMake can be updated ( if needed ) to support it.

jaimergp:
I'm fine dropping the PR if it makes no sense anymore due to changes in the infrastructure. I guess we will not need the wrapper in the near future, it seems? 

There are some pieces of the PR that might be useful to merge though, like the `-ccbin` fixes. Thoughts?

jakirkham:
Since this PR was raised, we have added CUDA 12 packages including actual CUDA compiler packages (instead of this wrapper script), which are heavily tested with CMake

Meanwhile legacy CUDA support has fallen off to point that only CUDA 11.8 is still available

Am curious whether there is still interest in this testing for CUDA 11.8. If so, we could clean this up and get it in. If not, maybe we should close to refocus on other efforts

jaimergp:
I won't be working on this so I'll close, but feel free to repurpose the diff in another PR. Thanks!

jakirkham:
Sounds reasonable. Thanks again for working on this Jaime! 🙏

If you see something missing in our testing in the `cuda-nvcc*` feedstocks, please let us know and we can work with you to incorporate your feedback 🙂

PR #63, opened 2021-01-26
Add cuda 11.1 and 11.2

jaimergp:
Checklist

Supersedes #50 

- [ ] Blocked by discussion at https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1162#issuecomment-776652909


recipe/meta.yaml
```diff
@@ -20,25 +32,33 @@ outputs:
...
     requirements:
       build:      # [win]
         - m2-sed  # [win]
       host:
         # Needed to symlink libcuda into sysroot libs.
         - {{ compiler("c") }}
-        - sysroot_linux-64 2.17  # [linux and cuda_compiler_version == "11.0"]
-        - __glibc >=2.17         # [linux and cuda_compiler_version == "11.0"]
+{% if cuda_major_minor >= (11, 0) %}
+        - sysroot_linux-64 2.17   # [linux]
```
leofang, line 45: I thought this should be in the `build` section, not `host`?
https://conda-forge.org/docs/maintainer/knowledge_base.html?highlight=sysroot#using-centos-7

recipe/meta.yaml
```diff
@@ -20,25 +32,33 @@ outputs:
...
+{% endif %}
       run:
-        - sed                    # [linux]
-        - sysroot_linux-64 2.17  # [linux and cuda_compiler_version == "11.0"]
-        - __glibc >=2.17         # [linux and cuda_compiler_version == "11.0"]
+        - sed                     # [linux]
+{% if cuda_major_minor >= (11, 0) %}
+        - sysroot_linux-64 2.17   # [linux]
+        - __glibc >=2.17          # [linux]
+{% endif %}
```
leofang, line 53: In cuDNN/cuTENSOR feedstocks, we put `__glibc` in `run_constrained`, not `run`, though I feel the effect should be the same given that you also added `run_exports`?
https://github.com/conda-forge/cutensor-feedstock/blob/cab81ab8c0ab8c3f4d37d05b6883094b3510585e/recipe/meta.yaml#L56-L58

recipe/meta.yaml
```diff
@@ -20,25 +32,33 @@ outputs:
...
     test:
       requires:
         - {{ compiler("c") }}
         # Host code is forwarded to a C++ compiler
         - {{ compiler("cxx") }}
-        - sysroot_linux-64 2.17  # [linux and cuda_compiler_version == "11.0"]
+{% if cuda_major_minor >= (11, 0) %}
+        - sysroot_linux-64 2.17   # [linux]
+{% endif %}
```
leofang, line 61: This is important.

leofang (commented):
The migration file looks OK to me, though I don't know how to kick off a migration process so I am not sure if this is the right place to do. I searched in the CF doc but it doesn't seem to be documented 🤔

One thing I am wondering, though, is that given we have to do a migration for CUDA 11.1/11.2 anyway, can't we just merge `.ci_support/migrations/windows_cuda111_112.yaml` into `.ci_support/migrations/cuda111_112.yaml` and use a selector like `# [linux64 or win]` to do the job? I think it's doable: https://github.com/conda-forge/cutensor-feedstock/blob/master/recipe/conda_build_config.yaml Just curious.

recipe/meta.yaml
jaimergp: The restraints [were already configured like this in the recipe](https://github.com/conda-forge/nvcc-feedstock/blob/master/recipe/meta.yaml#L12-L35). I only changed from selector syntax to Jinja conditionals because I found no other way of say `cuda_major >= 11` (see commits at #50 for a collection of failed attempts).

jaimergp:
> The migration file looks OK to me, though I don't know how to kick off a migration process so I am not sure if this is the right place to do.

I am bringing over the files submitted to https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1162. Docs drafting this process are available at https://conda-forge.org/docs/maintainer/knowledge_base.html#adding-support-for-a-new-cuda-version. It involves manual migrations to create the CUDA infrastructure first.

jaimergp:
Note that 

> One thing I am wondering, though, is that given we have to do a migration for CUDA 11.1/11.2 anyway, can't we just merge `.ci_support/migrations/windows_cuda111_112.yaml` into `.ci_support/migrations/cuda111_112.yaml` and use a selector like `# [linux64 or win]` to do the job?

Yep, this was suggested in that other PR and will bring the changes here once that's accepted!

jaimergp:
I merged the migrators and re-rendered, which made no changes, which is a good thing :)

recipe/meta.yaml
leofang: Yeah I did notice the constraints were added in `run_exports`. What I meant is it seems conda expects `__glibc`, which is a virtual package, to be added in `run_constrained`. At least this is what's documented:
https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#run-constrained
It seems there might be some subtitles I do not fully understand 😅

leofang:
> I am bringing over the files submitted to [conda-forge/conda-forge-pinning-feedstock#1162](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1162). Docs drafting this process are available at https://conda-forge.org/docs/maintainer/knowledge_base.html#adding-support-for-a-new-cuda-version. It involves manual migrations to create the CUDA infrastructure first.

Ah right, thanks for reminding me Jaime! I was searching for documentation on a general migration procedure but missed this one (which is all I want to know) 😅

jaimergp:
This is all passing now, given that we have the necessary bits in place (Docker images, Windows installer, cudatoolkits).

jaimergp:
Note that 10.2 is getting the Docker image for 9.2 for some reason. I need to expand the matrix again using a custom `conda_build_config.yaml` too, but my local efforts so far have not succeeded, so I haven't pushed anything yet.

```yaml
# When adding or removing cuda versions, make sure that the following entries are "zipped";
# e.g. each entry in cuda_compiler_version must have a matching entry in the other keys,
# considering the effect of the selector:
#   cuda_compiler_version
#   cudnn
#   cdt_name
#   docker_image

cuda_compiler_version:
  - None  # this is for cpu-only builds

  - 9.2    # [linux64]
  - 10.0   # [linux64 or win]
  - 10.1   # [linux64 or win]
  - 10.2   # [linux64 or win]
  - 11.0   # [linux64 or win]
  - 11.1   # [linux64 or win]
  - 11.2   # [linux64 or win]

docker_image:                                   # [os.environ.get("BUILD_PLATFORM", "").startswith("linux-")]
  # start cuda_compiler_version == None
  - quay.io/condaforge/linux-anvil-comp7        # [os.environ.get("BUILD_PLATFORM") == "linux-64"]
  - quay.io/condaforge/linux-anvil-aarch64      # [os.environ.get("BUILD_PLATFORM") == "linux-aarch64"]
  - quay.io/condaforge/linux-anvil-ppc64le      # [os.environ.get("BUILD_PLATFORM") == "linux-ppc64le"]
  - quay.io/condaforge/linux-anvil-armv7l       # [os.environ.get("BUILD_PLATFORM") == "linux-armv7l"]
  # end of cuda_compiler_version == None

  - quay.io/condaforge/linux-anvil-cuda:9.2     # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
  - quay.io/condaforge/linux-anvil-cuda:10.0    # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
  - quay.io/condaforge/linux-anvil-cuda:10.1    # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
  - quay.io/condaforge/linux-anvil-cuda:10.2    # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
  - quay.io/condaforge/linux-anvil-cuda:11.0    # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
  - quay.io/condaforge/linux-anvil-cuda:11.1    # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
  - quay.io/condaforge/linux-anvil-cuda:11.2    # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]

cudnn:
  # start cuda_compiler_version == None
  - undefined
  # end cuda_compiler_version == None

  - 7      # [linux64]                  # CUDA 9.2
  - 7      # [linux64 or win]           # CUDA 10.0
  - 7      # [linux64 or win]           # CUDA 10.1
  - 7      # [linux64 or win]           # CUDA 10.2
  - 8      # [linux64 or win]           # CUDA 11.0
  - 8      # [linux64 or win]           # CUDA 11.1
  - 8      # [linux64 or win]           # CUDA 11.2

cdt_name:  # [linux]
  # start cuda_compiler_version == None
  - cos6   # [linux64]
  - cos7   # [linux and aarch64]
  - cos7   # [linux and ppc64le]
  - cos7   # [linux and armv7l]
  # end cuda_compiler_version == None

  - cos6   # [linux64]                  # CUDA 9.2
  - cos6   # [linux64]                  # CUDA 10.0
  - cos6   # [linux64]                  # CUDA 10.1
  - cos6   # [linux64]                  # CUDA 10.2
  - cos7   # [linux64]                  # CUDA 11.0
  - cos7   # [linux64]                  # CUDA 11.1
  - cos7   # [linux64]                  # CUDA 11.2

```

leofang:
@jaimergp CUDA 11.2.1 is out...https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#cuda-whats-new

jaimergp:
... yay? :D Now we have to add the new versions to `cudatoolkit-feedstock`, `docker-images`, `conda-forge-ci-setup` too. I'll add to [the tracking issue](https://github.com/conda-forge/cudatoolkit-feedstock/issues/47).

leofang:
I suppose we could skip 11.2.0 and just use 11.2.1, thoughts?

jaimergp:
But the packages are [already out](https://anaconda.org/conda-forge/cudatoolkit/files?version=11.2.0)...

leofang:
Yeah well, let's wait for 11.3 then? Given the unlikelyhood of spanning the build matrix...

jaimergp:
I think we can have `11.2.1` just fine. `conda` will get the latest one within `major.minor`. The only problem is that we might need to change how `cudatoolkit` handles patch versions.

leofang:
I think you're referring to runtime dependency, but there is also build time dependency to address. For example, cuSPARSE got significantly expanded and improved in 11.2.1, which is otherwise invisible to applications compiled with CUDA <= 11.2.0, so those applications (I am unaware of any, so just hypothetically) would need to be built with CUDA 11.2.1 to take advantage of new features.

jaimergp:
Build time deps would be provided through docker-images (Linux) and conda-forge-ci-setup (Windows). In that case, can we replace 11.2.0 with 11.2.1?

h-vetinari:
Also, we have pretty complete control over the build dependencies, right**? I mean, the migrator hasn't even started yet. So as long as 11.2.1 is available at the time of the migration, the improvements @leofang speaks of should already be picked up?

** E.g., is it possible to add a run-export for `cudatoolkit >=11.2.1,<11.3` to the 11.2 migrator? That way, the packages would be built with 11.2.1, and installing a package built with that would definitely upgrade whatever `cudatoolkits =11.2.0` are already floating around.

leofang:
But @jaimergp @h-vetinari aren't we proposing effectively the same idea that we just use 11.2.1 throughout the CF infrastructure? What's the difference between your replies and my "skipping 11.2.0 and jumping to 11.2.1"?

jaimergp:
Yes, I think we are in time to skip 11.2.0 being used for builds. We only need to update the build-time bits in the repos mentioned at https://github.com/conda-forge/cudatoolkit-feedstock/issues/47. 

The only tricky part might happen at `cudatoolkit-feedstock` where we might need to overwrite the [dict entry for 11.2](https://github.com/conda-forge/cudatoolkit-feedstock/blob/master/recipe/meta.yaml#L158) so it points to 11.2.1, not 11.2.0, which will imply that 11.2.0 would not receive further updates done in the recipe.  The alternative is to refactor a bit so the dict key is not `major.minor` but optionally `major.minor.patch`. Not a big deal, just something to take into account in the PR.

h-vetinari:
> What's the difference between your replies and my "skipping 11.2.0 and jumping to 11.2.1"?

No difference worth mentioning, just a bit of care around what's already live.

> The only tricky part might happen at `cudatoolkit-feedstock` where we might need to overwrite the [dict entry for 11.2](https://github.com/conda-forge/cudatoolkit-feedstock/blob/master/recipe/meta.yaml#L158) so it points to 11.2.1, not 11.2.0, which will imply that 11.2.0 would not receive further updates done in the recipe.

I think that would be completely fine.

isuruf:
@conda-forge-admin, rerender

jaimergp:
We don't want to merge this without building all the variants, right? We have to expand the build matrix for all versions.

jakirkham:
Thanks Jaime! 😀

PR #66, opened 2021-04-19
Add PowerPC

### LINKED ISSUE #21: PowerPC builds
It would be nice to add support for `ppc64le` architectures. I don't know if the current system would work OK for this situation, since it assumes one CUDA version = one Docker image.

Even if that gets solved, there are some external barriers:

* [Nvidia has Docker images](https://hub.docker.com/r/nvidia/cuda-ppc64le/), but no CentOS 6; only CentOS 7 and Ubuntu 16/18. I read at some point that `conda-forge` would eventually migrate to CentOS 7, but I have no idea when that's going to happen.
* [Anaconda has `cudatoolkit` for `ppc64le`](https://anaconda.org/anaconda/cudatoolkit), but only for `v9.0`.

I'll leave this here for future reference and update as needed. Thanks!

jaimergp:
Checklist
* [X] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)

Questions:
* Are the compiler versions ok?
* How do we need to handle the `sysroot-*` dependencies?

Closes #21


recipe/conda_build_config.yaml
```diff
@@ -6,15 +7,24 @@
 #   cdt_name
 #   docker_image
 
+zip_keys:
+  - cuda_compiler_version   # [linux64 or (linux and ppc64le)]
+  - cdt_name                # [linux64 or (linux and ppc64le)]
+  - docker_image            # [linux64 or (linux and ppc64le)]
+  - cudnn                   # [linux64]
```
isuruf, line 14: why is this not zipped with `ppc64le`?

recipe/conda_build_config.yaml
```diff
@@ -61,3 +75,7 @@ cdt_name:  # [linux]
   - cos7   # [linux64]                  # CUDA 11.0
   - cos7   # [linux64]                  # CUDA 11.1
   - cos7   # [linux64]                  # CUDA 11.2
+
+  - cos7   # [linux and ppc64le]        # CUDA 10.2
+  # - cos7   # [linux and ppc64le]        # CUDA 11.0
```
isuruf, line 80: Why not add these? If we are going to add a migrator, we should add these two. Otherwise we'll have to have 2 migrators.

jaimergp: I left those there as a reminder while I was submitting the PR to `docker-images`. They are now added.

jaimergp:
Ok, that did the trick. Now we need to wait for the `docker-images` PR.

recipe/conda_build_config.yaml
jaimergp: Fixed!

recipe/conda_build_config.yaml
```diff
@@ -46,6 +70,11 @@ cudnn:
   - 8      # [linux64 or win]           # CUDA 11.1
   - 8      # [linux64 or win]           # CUDA 11.2
 
+  - undefined   # [linux and ppc64le]        # CUDA 10.2
+  - undefined   # [linux and ppc64le]        # CUDA 11.0
+  - undefined   # [linux and ppc64le]        # CUDA 11.1
+  - undefined   # [linux and ppc64le]        # CUDA 11.2
```
kkraus14, line 76: We should probably package these in the future

jaimergp: Created an issue there so we don't forget!

jaimergp:
The Docker images are now available and the packages can get built!

A question is still pending though: what do we do about the [`sysroot` dependency](https://github.com/jaimergp/nvcc-feedstock/blob/ppc/recipe/meta.yaml#L45)? [PowerPC CUDA 11.0+ uses CentOS 8](https://github.com/conda-forge/docker-images/blob/c9cc8cf584da51dbda088266997c96761b8d1475/.github/workflows/ci.yaml#L112-L125), so I guess that requires a new sysroot or am I mistaken? @isuruf @kkraus14 @jakirkham

jakirkham:
^ @beckermr do you know what we need to do to add CentOS 8 support?

beckermr:
We have no support for centos8.

jaimergp:
That means we are stuck with CUDA 10.2 on `ppc64le` for now (I'm ok with that).

beckermr:
CentOS 8 is reaching EOL very early compared to 7. I am wondering if maybe upstream can keep to centos 7?

@jakirkham ?

jakirkham:
Unfortunately there are no CUDA Toolkit versions for CentOS 7 and CUDA 11.0+ on PowerPC or ARM. Only CentOS 8

cc @kkraus14 (in case I'm missing anything here)

kkraus14:
As far as I know the only supported platform is RHEL 8 for CUDA 11.0+. That being said, the installers are all based on the runfile, which typically aren't OS specific in any way. For PPC64 that is https://developer.download.nvidia.com/compute/cuda/11.3.0/local_installers/cuda_11.3.0_465.19.01_linux_ppc64le.run for CUDA 11.3.

We could possibly check the glibc versioning in there and see?

beckermr:
It is technically feasible to support cos8 in conda-forge, but the expected lifetime of that distribution is small because they shortened the support period. Thus I'd prefer to not spend time on that. Maybe @isuruf has thoughts?

beckermr:
Another comment: We should support newer glibc's than cos7, but we might want to get them some other place (e.g. rocky linux).

jaimergp:
> That being said, the installers are all based on the runfile, which typically aren't OS specific in any way

The [Docker images are based on COS8](https://github.com/conda-forge/nvcc-feedstock/pull/66), though. 

_And 11.3 is also available now 🤦_

jakirkham:
Going on a little bit of a tangent, but do we know to what extent other things like CDTs and yum installs will change when moving to Rocky?

isuruf:
> The Docker images are based on COS8, though.

We can just install the runfile on centos 7 docker image. I checked and cudatoolkit=11.0.3 conda ppc64le package only uses glibc_2.17 which is centos 7.

beckermr:
> Going on a little bit of a tangent, but do we know to what extent other things like CDTs and yum installs will change when moving to Rocky?

I have no idea. We'll want to confer with anaconda as well since they might choose a different solution and that could matter.

cc @chenghlee

isuruf:
Can we please move the conversation about centos 8 or other distro to another issue?

Let's keep the discussion here about whether we need centos 8 or not for cuda.

jaimergp:
What is the estimated effort for the two feasible options now?

* Package `sysroot` for COS8.
* Frankenstein Docker images for CUDA 11. 
  * What happens if CUDA12 jumps to COS9?

jakirkham:
Opened issue ( https://github.com/conda-forge/conda-forge.github.io/issues/1432 ) to discuss how to handle CentOS 8

isuruf:
> Package sysroot for COS8.

A few weeks probably and would have to fix some fires while we get it right.

> Frankenstein Docker images for CUDA 11.

A day

> What happens if CUDA12 jumps to COS9?

There's no COS9

jaimergp:
I see, thanks for the answers!

In the meantime, would it make sense to merge only for CUDA 10.2 while we figure out how to support CUDA 11+? I wouldn't even add the migrator yet, but some packages in "manual CUDA mode" (like OpenMM) could benefit from having the infrastructure pieces ready and build some packages with the stable bits.

jakirkham:
Makes sense to me :)

kkraus14 (approved):

jaimergp:
@conda-forge/nvcc everybody ok with merging this for 10.2 only?

jakirkham (approved):

jakirkham:
Thanks Jaime! 😄

PR #71, opened 2021-09-27
Use CUDA Enhanced Compatibility for 11.2+

jakirkham:
Leverages [CUDA Enhanced Compatibility]( https://docs.nvidia.com/deploy/cuda-compatibility/ ) for 11.2+. This should simplify updating to new CUDA versions going forward and cutdown on the maintenance burden of CUDA versions.

Checklist
* [x] Reset the build number to `0` (if the version changed)

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "nvcc" %}
-{% set number = 14 %}
+{% set number = 15 %}
 
 {% if cuda_compiler_version in (None, "None", True, False) %}
 {% set cuda_major = 0 %}


kkraus14:
I think we just need to make sure to loosen the `__cuda` constraint on the `cudatoolkit` package for 11.2+ as well moving forward.

leofang:
> I think we just need to make sure to loosen the `__cuda` constraint on the `cudatoolkit` package for 11.2+ as well moving forward.

I doubt this is safe. At least for applications depending on PTX JIT (such as NVRTC), the driver version must be compatible with the runtime CTK version. If it's too old then we'd see driver errors such as `CUDA_ERROR_UNSUPPORTED_PTX_VERSION`. This is best explained [here](https://docs.nvidia.com/cuda/nvrtc/index.html#nvrtc-library-versioning):
> However, the more recent NVRTC library may generate PTX with a version that is not accepted by the CUDA Driver API functions of an older CUDA driver, as explained in the CUDA Compatibility document. Some approaches to resolving this issue: ...

and [here](https://docs.nvidia.com/deploy/cuda-compatibility/index.html#application-considerations):
> Applications that compile device code to PTX will not work on older drivers. If the application requires PTX then admins have to upgrade the installed driver.

In the case of CuPy, we work around this limitation by always generating SASS (CUBIN) instead of PTX for CUDA 11.1+, as instructed in the NVRTC documentation. But I don't know about other CUDA packages on conda-forge.

kkraus14:
> I doubt this is safe. At least for applications depending on PTX JIT (such as NVRTC), the driver version must be compatible with the runtime CTK version. If it's too old then we'd see driver errors such as `CUDA_ERROR_UNSUPPORTED_PTX_VERSION`.

Yes, we're aware that there's potential landmines but there was some discussion and I believe a decision was made that the best bet for now would be to set `__cuda>=11.2,<12` as an export and if a project is using newer / incompatible features then they can override that export.

Once we move to individual packages as opposed to one big `cudatoolkit` package then we could consider revisiting the `__cuda` export constraint on the `libnvrtc` package?

recipe/meta.yaml
```diff
@@ -31,7 +31,11 @@ outputs:
         - libgcc-ng
       run_exports:
         strong:
+{% if cuda_major_minor < (11, 2) %}
```
leofang, line 34: I am probably too tired, but shouldn't this be `(11, 2) <= cuda_major_minor < (12, 0)`?

jakirkham: Ah had the cases flipped. Should be fixed now. Thanks! 😄

isuruf:
@kkraus14, even though we discussed this issue, the issue about PTX never came up in the discussion. In that case, I don't think it is safe to build with 11.4 and running with 11.2.

kkraus14:
Fair enough. It's a bit of a shame about the PTX versioning.

Maybe if we can have a virtual package related to GPU architecture in the future it would open some new opportunities here.

jakirkham:
I think only Numba uses PTX and it only uses that at runtime. Am not aware of Numba or other packages distributing PTX

leofang:
My take is to push this through as a way for libraries to take advantage of CUDA Enhanced Compatibility. I am not saying this as an NVIDIA employee, but rather as a GPU developer (for CuPy). This is the trajectory that we've gone through, and I expect there will be more and more users asking each library to adopt, and eventually it'd no longer be a concern for us.

jakirkham:
Agreed. I think we are at a similar point with CUDA support as we were with Windows UCRT several years back. If we go ahead and embrace it warts and all, this will massively simplify our maintenance burden going forward and the rough edges will get smoothed in later iterations.

kkraus14:
> I think only Numba uses PTX and it only uses that at runtime. Am not aware of Numba or other packages distributing PTX. So I don't think PTX is an issue for us

Any library that is compiled with PTX in it can potentially cause issues. I.E. RAPIDS libraries have PTX for the latest supported architecture: https://github.com/rapidsai/rapids-cmake/blob/branch-21.12/rapids-cmake/cuda/set_architectures.cmake#L77-L81

This means with an older driver if someone tries to run on an older card that doesn't have the SASS already compiled that it would error in trying to JIT the PTX.

jakirkham:
@raydouglass, do you have any thoughts on Keith's observation above?

leofang:
From the cmake file that Keith quoted, it looks like for each arch a SASS is generated, and in addition a PTX is generated for the latest arch (potentially for early adoption when a future arch is out). 

For this particular use case, my understanding (I think this is a public info) is if we have a new arch out (say `sm_90`), the CTK major version needs to be bumped anyway, so we should be safe with regard to the Enhanced Compatibility which is not guaranteed across major versions.

I am interested to learn other PTX use cases in the wild.

kkraus14:
Yea it's moreso that say someone tries to run this on Maxwell architecture then it would try to JIT the PTX and would error from the PTX version being too new for the driver version in this case.

jakirkham:
So it sounds like we are on the same page. Would it make sense to return to the individual PRs at this point to make sure they have the needed changes and then working on getting them in? Or are there other things we still need to discuss here before doing that?

kkraus14:
I think the only point of contention left is if we move forward in adding CUDA 11.3 / 11.4 docker images. I think we should move forward with building those images to enable packages that can leverage newer features to opt into using them.

Otherwise I think we have everything else in place.

jakirkham:
Yeah am ok with doing that. Currently the Docker images are waiting for the `cudatoolkit` packages to be produced ( https://github.com/conda-forge/cudatoolkit-feedstock/pull/60 ). So getting that in would help

leofang:
I thought we could do the opposite: Get rid of all CUDA images and use vanilla Linux images + CUDA conda packages to build? (As I demo'd in https://github.com/conda-forge/cupy-feedstock/pull/143)

jakirkham:
Think you are thinking about something else ( https://github.com/conda-forge/cudatoolkit-feedstock/issues/62 ). At some point yes, but we are not ready to go there yet.

Should add would like to keep the scope tight here to make sure we achieve this step in a reasonable timeline 😉

leofang:
Ah OK, wasn't aware of the timeline consideration. Sounds good to me.

jakirkham:
Have submitted PR ( https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/172 ) to hot-fix the CUDA 11.2 packages to relax their pinnings.

If folks have time to review that PR and this one, that would be appreciated 😄

kkraus14 (approved):
LGTM pending https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/172

recipe/meta.yaml
```diff
@@ -33,7 +33,11 @@ outputs:
...
           - cudatoolkit {{ cuda_compiler_version }}|{{ cuda_compiler_version }}.*
+{% else %}
+          - cudatoolkit >={{ cuda_compiler_version }},<{{ cuda_major + 1 }}
```
leofang, line 39: Shouldn't we spell 11.2 out explicitly? What if `cuda_compiler_version` is 11.3? Or, based on the current plan, we don't release nvcc for 11.3+?
```suggestion
          - cudatoolkit >=11.2,<{{ cuda_major + 1 }}
```

jakirkham: No because if we build with a newer `cudatoolkit` version the package may pull in newer features (symbols) that would be unavailable in previous versions. So we would still want to set the minimum to be the version that we built with

jakirkham:
Any other thoughts on this one? Also the repodata patch has been merged

leofang (approved):

PR #84, opened 2022-05-25
Symlink libcuda.so

galipremsagar:
Checklist


recipe/install_nvcc.sh
```diff
@@ -102,7 +102,12 @@ then
   LIBCUDA_SO_CONDA_NVCC_BACKUP="\${CONDA_BUILD_SYSROOT}/lib/libcuda.so-conda-nvcc-backup"
   mv "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
 fi
-ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so"
+if [[ "${CONDA_BUILD}" == 1 ]]
+then
+  ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so"
+else
+  ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_PREFIX}/lib/libcuda.so"
+fi
```
kkraus14, line 110: It would be nice to handle this outside of conda-build as well.

We could use the output of `gcc -print-sysroot` in order to reliably get the sysroot.

kkraus14 (changes requested):

jakirkham:
Would [enable `azure/store_build_artifacts` in `conda-forge.yml`]( https://conda-forge.org/docs/maintainer/conda_forge_yml.html?highlight=store_build_artifacts#azure-config ) to ensure binaries built here are available on Azure to try out.

Also would bump [the build number]( https://github.com/conda-forge/nvcc-feedstock/blob/6aa4413734dc43299bb948d561073e28d1f8d5e7/recipe/meta.yaml#L2 )

recipe/install_nvcc.sh
jakirkham: We are just exploring options here at this point.

That said, some folks are running into build issues with CMake locally and this one attempt to address that. It's possible we need to do something different in the end

jakirkham:
cc @robertmaynard

recipe/install_nvcc.sh
kkraus14: > That is a different location compared to CONDA_BUILD_SYSROOT, but it is searched by CMake

Where is `CONDA_BUILD_SYSROOT` during a conda build? I figured it would be in the `build` environment at a similar place to where `gcc -print-sysroot` points to.

> Would ${CONDA_PREFIX}/lib/stubs be safe?

I assume so. It would be nice to ship all of the cuda stub libraries in a package in addition to the driver stub.

galipremsagar:

conda-forge.yml
```diff
@@ -3,3 +3,4 @@ provider: {linux_ppc64le: azure, linux_aarch64: azure}
 github:
   branch_name: main
   tooling_branch_name: main
+  store_build_artifacts: True
```
jakirkham, line 6:
```suggestion
azure:
  store_build_artifacts: True
```

jakirkham:

recipe/install_nvcc.sh
robertmaynard: The locations outside of a conda build are:

* `CONDA_BUILD_SYSROOT` => `<env>/x86_64-conda-linux-gnu/sysroot`
* `gcc -print-sysroot` => `<env>/x86_64-conda-linux-gnu/sysroot`
* `CMAKE_COMPILER_SYSROOT` => `<env>/x86_64-conda-linux-gnu/sysroot/usr/`

We would want to place the stub either in `${CONDA_PREFIX}/lib/stubs` or in `$(gcc -print-sysroot)/usr` for it to be found by CMake.

jakirkham:
When extracting the `zip` file, a `build_artifacts` directory should show up. This can be treated as a channel. IOW one can do

```
conda install -c file:///<path to build_artifacts>/build_artifacts nvcc_linux-64
```

galipremsagar:
> When extracting the `zip` file, a `build_artifacts` directory should show up. This can be treated as a channel. IOW one can do
> 
> ```
> conda install -c file:///<path to build_artifacts>/build_artifacts nvcc_linux-64
> ```

I see, thanks. Didn't know this and was wondering why there is only `build_artifacts`.

jakirkham:
Yeah it is archiving the whole `build_artifacts` directory, which is what is used for the whole build. So contains more than just the package. If there are ways to improve the docs to make this clearer, would be interested on your feedback on this 🙂 Though maybe we can save for a PR after

recipe/install_nvcc.sh
kkraus14: Could it go into `${gcc -print-sysroot)/usr/lib`? That seems to make sense for me and mirrors where the driver is in a normal system install.

kkraus14: So then I think we can generally use `$(gcc -print-sysroot)/usr/lib` for this and ignore the `CONDA_BUILD_SYSROOT` environment variable entirely?

galipremsagar:

recipe/install_nvcc.sh
robertmaynard: Can we safely presume that `gcc` points to a conda gcc?

kkraus14: The way the current recipe is, I guess not 🙃.

Do we want to allow using this package with system compilers outside of the conda environment?

robertmaynard: How about we use `$(CONDA_BUILD_SYSROOT)/usr/lib` since that should resolve to the same location as `$(gcc -print-sysroot)/usr/lib`

kkraus14: `CONDA_BUILD_SYSROOT` is undefined within my environment so I don't think we can rely on that either.

robertmaynard: So our only option would be ${CONDA_PREFIX}/lib/stubs

kkraus14: Technically the package depends on the `sysroot_linux-64 2.17` package (https://github.com/conda-forge/nvcc-feedstock/blob/main/recipe/meta.yaml#L57) so a sysroot would be available in any conda environment with this package.

That being said, in theory someone could be using this package with host compilers (though I don't think we really support doing so), in which case it probably wouldn't search in the conda env sysroot, but would use the system sysroot which would presumably have the driver anyway.

kkraus14: Alternatively, NVIDIA upstreaming things properly based on https://github.com/conda-forge/cudatoolkit-feedstock/issues/62 would make all of this work a moot point.

Also alternatively, it's not declared in the EULA, but if redistributing the stub libraries was allowed then we could just add all of the stub libraries to the `cudatoolkit` package which would include the `libcuda.so` stub and make this a moot point as well.

jakirkham:
JFYI this seems to be roughly working for us in terms of building. Though the stub library can't stay where it is (as noted in the thread above). We also find that issue when loading libraries locally (though removing the stub library at runtime fixes that). Going to try to dig into this a bit more, but it may not be until tomorrow

vyasr:
Forgive my ignorance, but could someone please summarize exactly what use cases that we are trying to support?
- Can users make use of system host compilers, or is the assumption that they will be using conda-provided host compilers (gcc)?
- Does the build process _always_ use the libcuda.so stub, even if we're on a system that actually has GPU drivers enabled (i.e. if there is a functional libcuda.so available in addition to the stub)? Is that managed by the compiler's sysroot? Does the existence or nonexistence of a functional libcuda.so change what the build process might find?
- Is there any scenario, now or in the future where we would need to support an actual libcuda.so provided by conda? That is not the current state of the cudatoolkit conda-forge feedstock, but it may be the future, right?

robertmaynard:
> Forgive my ignorance, but could someone please summarize exactly what use cases that we are trying to support?
> 
> * Can users make use of system host compilers, or is the assumption that they will be using conda-provided host compilers (gcc)?

I would state that the goal is to have this file behave as is for users using system host compilers, with the updates allowing conda-provided compilers to find the libcuda.so stub in the conda env. Currently the assumption for system host compilers is that they can find an existing system `libcuda.so` ( stub or real ).

> * Does the build process _always_ use the libcuda.so stub, even if we're on a system that actually has GPU drivers enabled (i.e. if there is a functional libcuda.so available in addition to the stub)? Is that managed by the compiler's sysroot? Does the existence or nonexistence of a functional libcuda.so change what the build process might find?

CMake always searches for real libraries before stub libraries. Specifically CMake only searches inside `stubs/` after first searching other directories. Since we are adding a stub file to a system search location this will only be used by CMake as the last search location.

This behavior is managed by how CMake performs `find_library` calls and the order of search locations, which ensures that compiler / system extracted locations are the last location it searches.

> * Is there any scenario, now or in the future where we would need to support an actual libcuda.so provided by conda? That is not the current state of the cudatoolkit conda-forge feedstock, but it may be the future, right?

As far as I am aware this is something we will need to support. If we always place the `libcuda.so` in a `stubs` directory we won't clobber a real `libcuda.so` and will be future proofed.

recipe/install_nvcc.sh
```diff
@@ -102,7 +102,14 @@ then
...
   mv "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
 fi
-ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so"
+if [[ "\${CONDA_BUILD}" == 1 ]]
+then
+  echo "BUILDING"
+  ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so"
+else
+  echo "NON-BUILD"
+  ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_PREFIX}/lib/libcuda.so"
```
jakirkham, line 111: Would this work?

```suggestion
  ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_PREFIX}/lib/stubs/libcuda.so"
```

galipremsagar: When I activate the env, I get this error:
```
pgali@dt07:/nvme/0/pgali/cudf$ conda activate /nvme/0/pgali/envs/cudfdev
ln: failed to create symbolic link '/nvme/0/pgali/envs/cudfdev/lib/stubs/libcuda.so': No such file or directory
```

isuruf:
My question is: Why would anyone want to use this package in a non-conda-build environment?

recipe/install_nvcc.sh
jakirkham:
```suggestion
  mkdir -p "\${CONDA_PREFIX}/lib/stubs"
  ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_PREFIX}/lib/stubs/libcuda.so"
```

jakirkham:
Because it works better with the Conda compilers, which are also being used in build environments. That in turn because it useful to develop with the compilers used to build the packages later

isuruf:
If that's the case, then `CONDA_BUILD_SYSROOT` is defined and this PR is unnecessary.

kkraus14:
> If that's the case, then `CONDA_BUILD_SYSROOT` is defined and this PR is unnecessary.

That's not necessarily true. I install the `gcc` package but not the `c-compiler` package because I don't want the activation environment variable setting a bunch of compiler options automatically for me in my conda environment (as opposed to a conda build). `CONDA_BUILD_SYSROOT` is then not defined.

jakirkham:
Unclear. We were having issues with this and CMake. Hence the exploration here

recipe/install_nvcc.sh
galipremsagar: Thanks @jakirkham ! This works, was able to build and verify imports/tests.

recipe/install_nvcc.sh
```diff
@@ -102,7 +102,15 @@ then
   LIBCUDA_SO_CONDA_NVCC_BACKUP="\${CONDA_BUILD_SYSROOT}/lib/libcuda.so-conda-nvcc-backup"
   mv "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
 fi
-ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so"
+if [[ "\${CONDA_BUILD}" == 1 ]]
+then
+  echo "BUILDING"
```
kkraus14, line 107:
```suggestion
```

recipe/install_nvcc.sh
```diff
@@ -102,7 +102,15 @@ then
...
+  ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so"
+else
+  echo "NON-BUILD"
```
kkraus14, line 110:
```suggestion
```

conda-forge.yml
```diff
@@ -3,3 +3,5 @@ provider: {linux_ppc64le: azure, linux_aarch64: azure}
 github:
   branch_name: main
   tooling_branch_name: main
+azure:
+  store_build_artifacts: True
```
vyasr, line 7: I think we can remove this now that we have manually verified everything that we need, right?

recipe/install_nvcc.sh
```diff
@@ -91,18 +91,31 @@ export CMAKE_ARGS="\${CMAKE_ARGS}"
...
-  mv "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
+  # Make a backup of \$(libcuda.so) if it exists
+  if [[ -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" ]]
+  then
+    LIBCUDA_SO_CONDA_NVCC_BACKUP="\${CONDA_BUILD_SYSROOT}/lib/libcuda.so-conda-nvcc-backup"
+    mv "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
+  fi
+  ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so"
+else
+  # Make a backup of \$(libcuda.so) if it exists
```
kkraus14, line 111: Do we actually need to make a backup of this? I think the only thing we support here is this symlink and if someone is doing something else, they're on their own.

recipe/install_nvcc.sh
```diff
@@ -91,18 +91,31 @@ export CMAKE_ARGS="\${CMAKE_ARGS}"
...
+  if [[ -f "\${CONDA_PREFIX}/lib/stubs/libcuda.so" ]]
+  then
+    LIBCUDA_SO_CONDA_NVCC_BACKUP="\${CONDA_PREFIX}/lib/stubs/libcuda.so-conda-nvcc-backup"
+    mv "\${CONDA_PREFIX}/lib/stubs/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
+  fi
+  ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_PREFIX}/lib/stubs/libcuda.so"
```
kkraus14, line 117: If we're not doing the whole backup dance we'd probably want to make this a `ln -sf` to override an existing symlink.

recipe/install_nvcc.sh
```diff
@@ -91,18 +91,31 @@ export CMAKE_ARGS="\${CMAKE_ARGS}"
...
+if [[ "\${CONDA_BUILD}" == 1 ]]
 then
-  LIBCUDA_SO_CONDA_NVCC_BACKUP="\${CONDA_BUILD_SYSROOT}/lib/libcuda.so-conda-nvcc-backup"
-  mv "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
+  # Make a backup of \$(libcuda.so) if it exists
+  if [[ -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" ]]
+  then
+    LIBCUDA_SO_CONDA_NVCC_BACKUP="\${CONDA_BUILD_SYSROOT}/lib/libcuda.so-conda-nvcc-backup"
+    mv "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
+  fi
```
vyasr, line 108: Does this mean that even if a user has libcuda.so installed in their conda environment we will always ignore it and force a fallback to the stub pointing to the host libcuda.so? Based on @robertmaynard's reply [here](https://github.com/conda-forge/nvcc-feedstock/pull/84#issuecomment-1138672547) I thought we didn't want that?

vyasr: Just saw the conditional. So we're forcing use of the system libcuda.so in conda build environments, but not in user space. If a user installs libcuda.so into their own conda environment (not during conda build) then we do use that, just not during conda builds. Is that right?

jakirkham:
FWIW a few people have tested this locally and it seems to resolve the issue. After reviews are addressed, planning on getting this in. Though please let us know ASAP if there are issues with that

recipe/install_nvcc.sh
galipremsagar: So should we remove the backup for both cases or just when CONDA_BUILD==0?

galipremsagar: Removed taking a backup 👍

recipe/install_nvcc.sh
galipremsagar: Done

recipe/install_nvcc.sh
galipremsagar: > If a user installs libcuda.so into their own conda environment (not during conda build) then we do use that, just not during conda builds. Is that right?

From what @kkraus14 mentioned above, if they have the `.so` in this location already then they are on their own. We force symlink currently to the host (or) whatever the `CUDA_HOME` points to.

galipremsagar:

galipremsagar:

conda-forge.yml
galipremsagar, line 7:
```suggestion
```

galipremsagar:
Completed testing this fix locally and verified it works. @jakirkham | @kkraus14 could either one of you review when you have time?

conda-forge.yml
galipremsagar: Rolled back this change.

galipremsagar:

leofang:
LGTM but I'll let John/Keith/Isuru merge 👍

recipe/install_nvcc.sh
```diff
@@ -91,18 +91,25 @@ export CMAKE_ARGS="\${CMAKE_ARGS}"
...
+  ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so"
+else
+  ln -sf "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_PREFIX}/lib/stubs/libcuda.so"
```
isuruf, line 111: `lib/stubs` is specific to cmake. What about other build systems?

robertmaynard: I agree with @kkraus14  we should think of this as matching the upstream cuda toolkit layout for empty/stub libraries. 

Since the cuda toolkit provides a `stubs` subdirectory, CMake supports searching it when it can't find a real library first.

isuruf: The point of this whole package is to do away with cudatoolkit weirdness by making it standard for the compiler. (just like what other distros do)

kkraus14: Well the standard behavior would be for this stub package to live in `lib/stubs` and for the actual driver library to live in the sysroot. Technically what we're doing now is wrong where we can end up in situations where the stub `libcuda.so` is a newer version than what's running on the system and symbols could technically fail to be found at library load time.

This means instead of what we're currently doing with the stub library that we'd need to detect if the system is actively running a driver and if it is to copy or symlink those driver libraries into the conda environment sysroot.

galipremsagar:

recipe/install_nvcc.sh
jakirkham: The linking to the stub libraries comes from [this recommendation]( https://github.com/NVIDIA/nvidia-docker/wiki/Frequently-Asked-Questions#how-do-i-link-against-driver-apis-at-build-time-eg-libcudaso-or-libnvidia-mlso ). Also this comes up a few times in issues ( for example https://github.com/NVIDIA/nvidia-docker/issues/374#issuecomment-298097944 )

recipe/install_nvcc.sh
```diff
@@ -91,18 +91,25 @@ export CMAKE_ARGS="\${CMAKE_ARGS}"
...
+
+if [[ "\${CONDA_BUILD}" == 1 ]]
 then
-  LIBCUDA_SO_CONDA_NVCC_BACKUP="\${CONDA_BUILD_SYSROOT}/lib/libcuda.so-conda-nvcc-backup"
-  mv "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
+  # Make a backup of \$(libcuda.so) if it exists
+  if [[ -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" ]]
+  then
+    LIBCUDA_SO_CONDA_NVCC_BACKUP="\${CONDA_BUILD_SYSROOT}/lib/libcuda.so-conda-nvcc-backup"
+    mv "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
```
jakirkham, line 107:
```suggestion
    mv -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
```

jakirkham, line 107:
```suggestion
    mv -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
```

recipe/install_nvcc.sh
jakirkham: Moved to `stubs`

jakirkham:
@kkraus14 and @isuruf, could you please take another look? It is worth noting the changes here are meant for use outside conda-forge. In particular this fixes some issues users ran into locally and also recently on RAPIDS CI. Am eager to get this in to fix those issues.

recipe/install_nvcc.sh
isuruf: > Well the standard behavior would be for this stub package to live in lib/stubs and for the actual driver library to live in the sysroot.

This is wrong. Actual driver library `libcuda.so.1` should live in ldconfig path where the dynamic loader can find it or in `$PREFIX/lib`.
It's fine to have `libcuda.so` in `$PREFIX/lib` since it's only used for linking and the SONAME is `libcuda.so.1` which is different from `libcuda.so`.

isuruf: If you still want to keep this in `lib/stubs`, then you should add a corresponding change to `LDFLAGS` to look at `lib/stubs`.

isuruf: Or set `LIBRARY_PATH` to `lib/stubs`.

jakirkham: Sure. Would this work?

Guessing we don't want the `-Wl,-rpath,...` piece, but please let me know if I've missed something there

```suggestion
  ln -sf "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_PREFIX}/lib/stubs/libcuda.so"
  export LDFLAGS="${LDFLAGS} -L${CONDA_PREFIX}/lib/stubs"
```

isuruf: Yes, but you need to restore it in deactivate.sh

recipe/install_nvcc.sh
```diff
@@ -91,18 +91,25 @@ export CMAKE_ARGS="\${CMAKE_ARGS}"
...
+mkdir -p "\${CONDA_PREFIX}/lib/stubs"
 
 # Add \$(libcuda.so) shared object stub to the compiler sysroot.
 # Needed for things that want to link to \$(libcuda.so).
 # Stub is used to avoid getting driver code linked into binaries.
 
-# Make a backup of \$(libcuda.so) if it exists
-if [[ -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" ]]
+
+if [[ "\${CONDA_BUILD}" == 1 ]]
```
isuruf, line 101: Instead of conda-build, do this if `CONDA_BUILD_SYSROOT` is defined

recipe/install_nvcc.sh
jakirkham: That makes sense. How would this play with the `compilers`, which is doing some of this same saving itself?

isuruf: If you go with my other suggestion, this code path will only be when the `compilers` package is not used.

recipe/install_nvcc.sh
```diff
@@ -91,18 +91,25 @@ export CMAKE_ARGS="\${CMAKE_ARGS}"
...
 then
-  LIBCUDA_SO_CONDA_NVCC_BACKUP="\${CONDA_BUILD_SYSROOT}/lib/libcuda.so-conda-nvcc-backup"
-  mv "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
+  # Make a backup of \$(libcuda.so) if it exists
+  if [[ -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" ]]
```
kkraus14, line 104: Why do we need to do this? When would this ever exist in an environment that wasn't placed by this script in a way that we should support with this script?

recipe/install_nvcc.sh
kkraus14: > This is wrong. Actual driver library `libcuda.so.1` should live in ldconfig path where the dynamic loader can find it or in `$PREFIX/lib`.

In a normal installation both `libcuda.so` and `libcuda.so.1` come from the driver and get put in `/lib/x86_64-linux-gnu/` which I believe is the system sysroot (as opposed to the conda sysroot). The stub `libcuda.so` comes from the toolkit and ends up in the `stubs` directory.

Outside of conda, if someone is building something that uses the CUDA driver library without a driver installed on the machine, it is their responsibility to either point their compiler to the stub library or use something like CMake that automatically handles it for them. CMake will try to find a driver installed and running on the system first, and if it cannot find it will then fall back to using the stubs. Automatically setting `LDFLAGS` or `LIBRARY_PATH` feels like this package overstepping its boundaries.

> It's fine to have `libcuda.so` in `$PREFIX/lib` since it's only used for linking and the SONAME is `libcuda.so.1` which is different from `libcuda.so`.

I've seen plenty of code that dlopens `libcuda.so` (as opposed to `libcuda.so.1`) to sidestep manylinux violations and we would potentially blow that code up if we did this.

jakirkham: Ah I see ( https://github.com/conda-forge/nvcc-feedstock/pull/84#discussion_r898285673 ). Yeah was thinking about this

recipe/install_nvcc.sh
galipremsagar: Updated to switch when `CONDA_BUILD_SYSROOT` is defined

recipe/install_nvcc.sh
galipremsagar: Discarded this check and let it take a backup in anyway.

recipe/install_nvcc.sh
```diff
@@ -91,18 +91,22 @@ export CMAKE_ARGS="\${CMAKE_ARGS}"
...
+mkdir -p "\${CONDA_PREFIX}/lib/stubs"
 
 # Add \$(libcuda.so) shared object stub to the compiler sysroot.
 # Needed for things that want to link to \$(libcuda.so).
 # Stub is used to avoid getting driver code linked into binaries.
 
-# Make a backup of \$(libcuda.so) if it exists
-if [[ -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" ]]
+
+if [[ "\${CONDA_BUILD_SYSROOT}" != "" ]]
```
jakirkham, line 101: Should catch whether it is defined

```suggestion
if [[ ! -z "\${CONDA_BUILD_SYSROOT+x}" ]]
```

recipe/install_nvcc.sh
```diff
@@ -91,18 +91,22 @@ export CMAKE_ARGS="\${CMAKE_ARGS}"
...
 then
+  # Make a backup of \$(libcuda.so)
   LIBCUDA_SO_CONDA_NVCC_BACKUP="\${CONDA_BUILD_SYSROOT}/lib/libcuda.so-conda-nvcc-backup"
-  mv "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
+  mv -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
```
jakirkham, line 105: Think this should fix the CI issue

```suggestion
  if [[ -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" ]]
  then
    mv -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
  fi
```

jakirkham (commented):
Had a few minor suggestions to address the CI issue

recipe/install_nvcc.sh
```diff
@@ -154,8 +161,14 @@ if [[ -f ""\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"" ]]
 then
   mv -f "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}" "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so"
 else
-  rm -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so"
+  if [[ "\${CONDA_BUILD_SYSROOT}" != "" ]]
```
galipremsagar, line 164:
```suggestion
  if [[ ! -z "\${CONDA_BUILD_SYSROOT+x}" ]]
```

recipe/install_nvcc.sh
```diff
@@ -91,18 +91,25 @@ export CMAKE_ARGS="\${CMAKE_ARGS}"
 ### /CMake configurations
 
 mkdir -p "\${CONDA_BUILD_SYSROOT}/lib"
```
jakirkham, line 93:
```suggestion
```

recipe/install_nvcc.sh
```diff
@@ -91,18 +91,25 @@ export CMAKE_ARGS="\${CMAKE_ARGS}"
...
 
 # Add \$(libcuda.so) shared object stub to the compiler sysroot.
 # Needed for things that want to link to \$(libcuda.so).
 # Stub is used to avoid getting driver code linked into binaries.
 
-# Make a backup of \$(libcuda.so) if it exists
-if [[ -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" ]]
+
+if [[ ! -z "\${CONDA_BUILD_SYSROOT+x}" ]]
 then
```
jakirkham, line 102:
```suggestion
then
  mkdir -p "\${CONDA_BUILD_SYSROOT}/lib"
```

jakirkham (commented):
Found a stray `CONDA_BUILD_SYSROOT` outside the branch checking it is defined. Suggesting moving into the branch

recipe/install_nvcc.sh
```diff
@@ -90,19 +90,26 @@ export CMAKE_ARGS="\${CMAKE_ARGS}"
 
 ### /CMake configurations
 
-mkdir -p "\${CONDA_BUILD_SYSROOT}/lib"
+mkdir -p "\${CONDA_PREFIX}/lib/stubs"
```
jakirkham, line 93:
```suggestion
```

recipe/install_nvcc.sh
```diff
@@ -90,19 +90,26 @@ export CMAKE_ARGS="\${CMAKE_ARGS}"
...
+  # Make a backup of \$(libcuda.so)
   LIBCUDA_SO_CONDA_NVCC_BACKUP="\${CONDA_BUILD_SYSROOT}/lib/libcuda.so-conda-nvcc-backup"
-  mv "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
+  if [[ -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" ]]
+  then
+    mv -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
+  fi
+  ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so"
+else
+  ln -sf "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_PREFIX}/lib/stubs/libcuda.so"
```
jakirkham, line 111:
```suggestion
  mkdir -p "\${CONDA_PREFIX}/lib/stubs"
  ln -sf "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_PREFIX}/lib/stubs/libcuda.so"
```

jakirkham (commented):
Suggesting moving `stubs` directory creation into relevant branch

jakirkham:

recipe/install_nvcc.sh
```diff
@@ -90,19 +90,24 @@ export CMAKE_ARGS="\${CMAKE_ARGS}"
...
 then
+  mkdir -p "\${CONDA_BUILD_SYSROOT}/lib"
+  # Make a backup of \$(libcuda.so)
   LIBCUDA_SO_CONDA_NVCC_BACKUP="\${CONDA_BUILD_SYSROOT}/lib/libcuda.so-conda-nvcc-backup"
-  mv "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
+  if [[ -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" ]]
+  then
+    mv -f "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so" "\${LIBCUDA_SO_CONDA_NVCC_BACKUP}"
+  fi
+  ln -s "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_BUILD_SYSROOT}/lib/libcuda.so"
```
kkraus14, line 106: I believe there's some OS package managers where the toolkit gets installed into the standard system library / include locations and I'm not sure this is the location of where the stubs end up. It would be nice to verify that this stub library exists and warn or error if it doesn't.

recipe/install_nvcc.sh
```diff
@@ -90,19 +90,24 @@ export CMAKE_ARGS="\${CMAKE_ARGS}"
...
+else
+  mkdir -p "\${CONDA_PREFIX}/lib/stubs"
+  ln -sf "\${CUDA_HOME}/lib64/stubs/libcuda.so" "\${CONDA_PREFIX}/lib/stubs/libcuda.so"
```
kkraus14, line 109: Here as well. Need to verify that the stub library exists.

kkraus14 (changes requested):

recipe/install_nvcc.sh
jakirkham: That makes sense.

Though it is worth noting we do [raise]( https://github.com/conda-forge/nvcc-feedstock/blob/570b94fae039161e3c043d2b255e293da351e31a/recipe/install_nvcc.sh#L60-L64 ) if `libcuda.so` is not found.

Also wouldn't it be possible for the user to override `$CUDA_HOME`? If the user doesn't do that, wouldn't [our `cuda-gdb` detection logic]( https://github.com/conda-forge/nvcc-feedstock/blob/570b94fae039161e3c043d2b255e293da351e31a/recipe/install_nvcc.sh#L44 ) find the right location (IOW does `cuda-gdb` also land in a system location when `libcuda.so` does)?

kkraus14 (approved):

recipe/install_nvcc.sh
jakirkham: Ah good point. Yeah happy to follow up on these deployment issues on a case-by-case basis as they come up

jakirkham:
Thanks all! 🙏

Looks like we resolved the remaining comments. So went ahead and merged. If it turns out we missed something, happy to follow up in a new issue/PR 🙂

PR #87, opened 2022-06-29
Remove version constraint of `nvcc` to match with host cuda version

galipremsagar:
This PR removes the version constraint that `nvcc` imposes to have the matching CUDA version on the host machine installed. Upon some searching, relaxing & testing this fix it appears that there is no version-specific file being shipped along with `nvcc` conda package, but I could be wrong here too. Hence requesting review from maintainers since this check has been present since its inception.

Rationale behind this PR is to allow users to utilized any version of `nvcc` package without upgrading their local host CUDA installations.

Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

### DIFF
--- a/recipe/install_nvcc.sh
+++ b/recipe/install_nvcc.sh
@@ -65,7 +65,13 @@ fi
 
 if [[ -z "\$(\${CUDA_HOME}/bin/nvcc --version | grep "Cuda compilation tools, release ${PKG_VERSION}")" ]]
 then
-  echo "Warning: Version of installed CUDA didn't match package"
+  if [[ ! -z "\${CONDA_BUILD_SYSROOT+x}" ]]
+  then
+    echo "Version of installed CUDA didn't match package"
+    return 1
+  else
+    echo "Warning: Version of installed CUDA didn't match package"
+  fi
 fi
 
 export CUDA_HOME="\${CUDA_HOME}"
--- a/recipe/install_nvcc.sh
+++ b/recipe/install_nvcc.sh
@@ -65,7 +65,7 @@ fi
 
 if [[ -z "\$(\${CUDA_HOME}/bin/nvcc --version | grep "Cuda compilation tools, release ${PKG_VERSION}")" ]]
 then
-  if [[ "\${CONDA_BUILD}"="1" ]]
+  if [[ "\${CONDA_BUILD}" = "1" ]]
   then
     echo "Version of installed CUDA didn't match package"
     return 1
--- a/recipe/install_nvcc.sh
+++ b/recipe/install_nvcc.sh
@@ -65,8 +65,7 @@ fi
 
 if [[ -z "\$(\${CUDA_HOME}/bin/nvcc --version | grep "Cuda compilation tools, release ${PKG_VERSION}")" ]]
 then
-  echo "Version of installed CUDA didn't match package"
-  return 1
+  echo "Warning: Version of installed CUDA didn't match package"
 fi
 
 export CUDA_HOME="\${CUDA_HOME}"
--- a/recipe/install_nvcc.sh
+++ b/recipe/install_nvcc.sh
@@ -65,7 +65,7 @@ fi
 
 if [[ -z "\$(\${CUDA_HOME}/bin/nvcc --version | grep "Cuda compilation tools, release ${PKG_VERSION}")" ]]
 then
-  if [[ ! -z "\${CONDA_BUILD+x}" ]]
+  if [[ "\${CONDA_BUILD}"="1" ]]
   then
     echo "Version of installed CUDA didn't match package"
     return 1
--- a/recipe/install_nvcc.sh
+++ b/recipe/install_nvcc.sh
@@ -65,7 +65,7 @@ fi
 
 if [[ -z "\$(\${CUDA_HOME}/bin/nvcc --version | grep "Cuda compilation tools, release ${PKG_VERSION}")" ]]
 then
-  if [[ ! -z "\${CONDA_BUILD_SYSROOT+x}" ]]
+  if [[ ! -z "\${CONDA_BUILD+x}" ]]
   then
     echo "Version of installed CUDA didn't match package"
     return 1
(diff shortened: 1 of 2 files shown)


galipremsagar:
Synced offline with @leofang and agreed on erroring only during conda build time and just raising a warning for end-user during conda activation.

recipe/install_nvcc.sh
```diff
@@ -65,8 +65,13 @@ fi
 
 if [[ -z "\$(\${CUDA_HOME}/bin/nvcc --version | grep "Cuda compilation tools, release ${PKG_VERSION}")" ]]
 then
-  echo "Version of installed CUDA didn't match package"
-  return 1
+  if [[ ! -z "\${CONDA_BUILD_SYSROOT+x}" ]]
```
leofang, line 68: Sorry I just woke up and am bad at negations, but do we need `!`?

leofang: Shouldn't it just be
```
if [[ -n "${CONDA_BUILD_SYSROOT}" ]]
```
?

leofang: I think the problem is in `+x`. Could you do a quick check in your terminal?

leofang: We want this if branch to be not hit if `CONDA_BUILD_SYSROOT` is not set.

leofang: Oh it's the backslash `\`...

galipremsagar: While looking into this I found that `CONDA_BUILD_SYSROOT` is being set when `sysroot_linux-64` is present in the conda environment. I was under an assumption this is only set during conda-build process and not in a user conda-environment. @isuruf is this an expected behavior?

leofang: I like `CONDA_BUILD_STATE`. We also used it in msmpi-feedstock to detect if we're using conda-build.

jakirkham: Could you please clarify how you would like to use `CONDA_BUILD_STATE` here, @jaimergp?

leofang: We want to preserve the current behavior (to fail if the version mismatches) but only inside the conda-build env (on conda-forge).

jakirkham: In that case it sounds like checking `CONDA_BUILD` would be sufficient

galipremsagar: Updated the script to use `CONDA_BUILD`

recipe/install_nvcc.sh
```diff
@@ -65,8 +65,13 @@ fi
 
 if [[ -z "\$(\${CUDA_HOME}/bin/nvcc --version | grep "Cuda compilation tools, release ${PKG_VERSION}")" ]]
 then
-  echo "Version of installed CUDA didn't match package"
-  return 1
+  if [[ ! -z "\${CONDA_BUILD+x}" ]]
```
jakirkham, line 68: According to [the docs]( https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#environment-variables-set-during-the-build-process ), this is always set to `1` when it is set. So we could just compare to that value

The current statement would evaluate to `true` if it is setting to anything (including "" and "0"). Not sure that anything actually uses those other values, but we may still want those to evaluate to `false` for good measure

galipremsagar: Thanks John, updated with a strict comparison

recipe/install_nvcc.sh
```diff
@@ -65,8 +65,13 @@ fi
 
 if [[ -z "\$(\${CUDA_HOME}/bin/nvcc --version | grep "Cuda compilation tools, release ${PKG_VERSION}")" ]]
 then
-  echo "Version of installed CUDA didn't match package"
-  return 1
+  if [[ "\${CONDA_BUILD}"="1" ]]
```
jakirkham, line 68:
```suggestion
  if [[ "\${CONDA_BUILD}" = "1" ]]
```

recipe/install_nvcc.sh
```diff
@@ -65,8 +65,13 @@ fi
 
 if [[ -z "\$(\${CUDA_HOME}/bin/nvcc --version | grep "Cuda compilation tools, release ${PKG_VERSION}")" ]]
 then
-  echo "Version of installed CUDA didn't match package"
-  return 1
+  if [[ "\${CONDA_BUILD}" = "1" ]]
+  then
+    echo "Version of installed CUDA didn't match package"
```
jakirkham, line 70:
```suggestion
    echo "Error: Version of installed CUDA didn't match package"
```

jakirkham:
Would also bump the build number here

https://github.com/conda-forge/nvcc-feedstock/blob/4acf86c13bfd9a828d3e44c1cd9e1255d0474f73/recipe/meta.yaml#L2

That should generate new packages with this change

galipremsagar:
> Would also bump the build number here
> 
> https://github.com/conda-forge/nvcc-feedstock/blob/4acf86c13bfd9a828d3e44c1cd9e1255d0474f73/recipe/meta.yaml#L2
> 
> That should generate new packages with this change

Done 👍

jakirkham (approved):

jakirkham:
Thanks Prem! 🙏 LGTM ✅

Will give others a chance to look 🙂

recipe/install_nvcc.sh
```diff
@@ -65,8 +65,13 @@ fi
 
 if [[ -z "\$(\${CUDA_HOME}/bin/nvcc --version | grep "Cuda compilation tools, release ${PKG_VERSION}")" ]]
 then
-  echo "Version of installed CUDA didn't match package"
-  return 1
+  if [[ "\${CONDA_BUILD}" = "1" ]]
```
leofang, line 68: Note to self: the extra backslash is needed because this is part of codegen, not the actual script.

leofang (approved):
Thanks, @galipremsagar @jakirkham. LGTM.

recipe/install_nvcc.sh
leofang: To be eliminated in #91.

leofang:
Should we port this change to the Windows script?

jakirkham:
Seems reasonable

PR #68, opened 2021-05-14
Standalone (de)activate scripts; Early exit when outside conda-build environments 

leofang:
Close #40, close #60, close #65.

This PR changes 3 things (following the commit order). Please pay particular attention to the 2nd one, as it would be a breaking change:
1. Make Linux activate/deactivate scripts standalone (which is already done for the Windows ones)
2. Prevent `nvcc` from being used outside conda-build envs (we call `rm -f` and we also rely on `CONDA_BUILD_SYSROOT`, so there is really not feasible for normal users to use)
3. Add myself as a maintianer

Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

PR #91, opened 2022-07-22
Make Linux (de)activate scripts standalone like on Windows

leofang:
Supersede #68. Close #68.

Checklist


leofang:

recipe/linux/install_nvcc.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -xeuo pipefail
+
+# Activation script
+mkdir -p "${PREFIX}/etc/conda/activate.d"
+sed -i "s/__PKG_VERSION__/$PKG_VERSION/g" linux/activate.sh
```
leofang, line 7:
```suggestion
sed -i "s/__PKG_VERSION__/$PKG_VERSION/g" $RECIPE_DIR/linux/activate.sh
```

jakirkham:
Appreciate the effort here Leo, but am not sure this is going to work.

We rely on having some of these environment variables getting replaced while leaving others to be determined at runtime. If we go this route, it may require some patching of the scripts, using `sed`, and/or injecting the full paths into the scripts for prefix replacement.

If we find something that works here, would be good to do a fair bit of testing to make sure everything is still behaving as expected.

leofang:
Hmmm I thought I did that, but maybe there're a few other variables that I missed. Please check the Windows scripts. The structure is identical now, so if we can do the variable injection on Windows, we should be able to do that on Linux too.

That said, I cannot continue today for some urgent tasks. Let me mark this as draft.

jakirkham:
Gotcha. Yeah can take a look later

No worries. Thanks for the effort here. Agree this would be more readable/maintainable with this change

recipe/install_nvcc.sh
```diff
@@ -1,189 +0,0 @@
...
-fi
-
-EOF
-
-# Create `nvcc` script in `bin` so it can be easily run.
-mkdir -p "${PREFIX}/bin"
-cat > "${PREFIX}/bin/nvcc" <<'EOF'
-#!/bin/bash
-for arg in "${@}" ; do
-  case ${arg} in -ccbin)
```
leofang, line 182: @conda-forge/nvcc Does anyone know why we had an unmatched parenthesis here?!

leofang: Oh nvm, I forgot bash `case` needs it...

leofang:
> We rely on having some of these environment variables getting replaced while leaving others to be determined at runtime. If we go this route, it may require some patching of the scripts, using `sed`, and/or injecting the full paths into the scripts for prefix replacement.

Hi @jakirkham it appears to me that we only need to worry about `PKG_VERSION` (which was already correctly handled). Am I missing something?

jakirkham:

conda-forge.yml
```diff
@@ -1,5 +1,8 @@
 conda_forge_output_validation: true
 provider: {linux_ppc64le: azure, linux_aarch64: azure}
+# remove the `azure` section and re-render before merging
+azure:
+  store_build_artifacts: False
```
leofang, line 5:
```suggestion
  store_build_artifacts: true
```

leofang:

conda-forge.yml
```diff
@@ -2,7 +2,7 @@ conda_forge_output_validation: true
 provider: {linux_ppc64le: azure, linux_aarch64: azure}
 # remove the `azure` section and re-render before merging
 azure:
-  store_build_artifacts: False
+  store_build_artifacts: true
```
jakirkham, line 5: Doh 🤦‍♂️

Thanks Leo 😅

jakirkham:
@galipremsagar @vyasr, could you please help test this when you have a chance?

Prebuilt Conda binaries can be found on CI (relevant [docs]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#downloading-prebuilt-packages-from-ci )). More context can also be found in issue ( https://github.com/conda-forge/conda-forge.github.io/issues/1424 ).

conda-forge.yml
leofang: No worries! But I do notice one thing odd: I checked with the past record, it takes 40-50 mins to rerender this feedstock. Is this expected? Is it because this is a compiler package?

galipremsagar:
> @galipremsagar @vyasr, could you please help test this when you have a chance?
> 
> Prebuilt Conda binaries can be found on CI (relevant [docs](https://conda-forge.org/docs/maintainer/updating_pkgs.html#downloading-prebuilt-packages-from-ci)). More context can also be found in issue ( [conda-forge/conda-forge.github.io#1424](https://github.com/conda-forge/conda-forge.github.io/issues/1424) ).

Verified locally with the artifacts and works on my machine 👍

recipe/linux/activate.sh
```diff
@@ -0,0 +1,110 @@
...
+
+if [[ ! -f "${CUDA_HOME}/lib64/stubs/libcuda.so" ]]
+then
+    echo "File ${CUDA_HOME}/lib64/stubs/libcuda.so doesn't exist"
+    return 1
+fi
+
+if [[ -z "$(${CUDA_HOME}/bin/nvcc --version | grep "Cuda compilation tools, release __PKG_VERSION__")" ]]
+then
+  if [[ "${CONDA_BUILD}" = "1" ]]
```
jaimergp, line 63:
```suggestion
  if [[ "${CONDA_BUILD}" == "1" ]]
```

Double brackets -> double equals :D

jaimergp:
@leofang Do you think it's worth it to use `set -euo pipefail` in the (de)activation scripts? I am not sure if you considered some variables might be undefined (like `CONDA_BUILD_SYSROOT`) under certain circumstances.

recipe/linux/activate.sh
leofang: Thanks, Jaime. In Bash it works fine:
```
$ CONDA_BUILD=1; if [[ "${CONDA_BUILD}" = "1" ]]; then echo "ok"; else echo "not ok"; fi
ok
$ CONDA_BUILD=1; if [[ "${CONDA_BUILD}" == "1" ]]; then echo "ok"; else echo "not ok"; fi
ok
$ CONDA_BUILD=0; if [[ "${CONDA_BUILD}" = "1" ]]; then echo "ok"; else echo "not ok"; fi
not ok
$ CONDA_BUILD=0; if [[ "${CONDA_BUILD}" == "1" ]]; then echo "ok"; else echo "not ok"; fi
not ok
```
But I like double equals better, too.

leofang:
> @leofang Do you think it's worth it to use `set -euo pipefail` in the (de)activation scripts? I am not sure if you considered some variables might be undefined (like `CONDA_BUILD_SYSROOT`) under certain circumstances.

I'd like to leave this question to @jakirkham @galipremsagar. In #87 and other PRs we tried to make the (de)activation script work outside of conda-build. My only concern is if the (de)activation script fails conda may be left in an invalid state. I seem to remember seeing a similar issue in the past but can't recall where.

leofang:
Another note is we currently have both `CONDA_BUILD_SYSROOT` and `CONDA_BUILD` used in the script. I aimed to do a simple migration in this PR so I didn't bother to touch them. Let me know if we need to unify the usage in this PR.

recipe/linux/activate.sh
leofang: (And by the way, the single equal is the status quo, not my change.)

recipe/linux/activate.sh
```diff
@@ -0,0 +1,110 @@
...
+
+if [[ ! -f "${CUDA_HOME}/lib64/stubs/libcuda.so" ]]
+then
+    echo "File ${CUDA_HOME}/lib64/stubs/libcuda.so doesn't exist"
+    return 1
+fi
+
+if [[ -z "$(${CUDA_HOME}/bin/nvcc --version | grep "Cuda compilation tools, release __PKG_VERSION__")" ]]
+then
+  if [[ "${CONDA_BUILD}" == "1" ]]
```
jakirkham, line 63: Sorry to push back on this, but the single equals was here because that is what [POSIX shells do]( https://stackoverflow.com/a/20449556 ) (`==` is unique to Bash). Would suggest we keep the POSIX compliant version

```suggestion
  if [[ "${CONDA_BUILD}" = "1" ]]
```

<br>
cc @jaimergp (for awareness)

recipe/linux/activate.sh
jakirkham: The `=` is a POSIX thing. So would suggest we stick with it ( https://github.com/conda-forge/nvcc-feedstock/pull/91#discussion_r931478978 )

recipe/linux/activate.sh
jaimergp: In that case shouldn't we use single brackets? Both work though, so no big deal.

jakirkham: Good point. Fixed above in an edit.

On a larger point, it would be nice to have shell agnostic activation/deactivation scripts ( https://github.com/conda/conda/issues/6820 ), but that is outside of the scope here.

leofang: Thanks for raising the discussion John. I wasn't as keen on the different shell conventions as I'd like to, so it's good that you caught it. If you could help we can try to make it POSIX complaint in this PR, but I am not knowledgeable enough (nor do I have the bandwidth) to do it myself 😅

jakirkham: We could add `shellcheck` to the test requirements and start running it on these scripts to check compliance

That said, I wouldn't worry about doing this in this PR. We can save it for future work ( https://github.com/conda-forge/nvcc-feedstock/issues/92 )

conda-forge.yml
jakirkham: Idk if I'd use the word "expected", but yes it does seem to take a long time to re-render this feedstock. Think that is a result of the rather large matrix of versions, architectures, etc. we include here. Issue ( https://github.com/conda-forge/conda-smithy/issues/1307 ) discusses long re-rendering cases and what options we might have to improve re-rendering time

leofang: Thanks for pointer!

vyasr:
@jakirkham @leofang @galipremsagar I was also able to verify that this worked for me locally.

jakirkham (approved):

jakirkham:
@jaimergp, could you please take another look when you have a chance? 🙂

leofang:

leofang:
Let's get this merged. Thanks everyone for reviewing & testing!

PR #106, opened 2024-08-16
Clean up; drop CUDA <11.8, add stdlib; avoid `libgcc-ng` in ignore_run_exports

### LINKED ISSUE #100: Drop CUDA<11.8 builds?
Noticed that CI still builds 9.2, 10.0, 10.1, 10.2, 11.0, 11.1, all of which have already been dropped by conda-forge. Given the number of CI jobs, I'm wondering if this is still useful/necessary?

CC @conda-forge/nvcc

### LINKED ISSUE #107: BUG: `nvcc_linux-aarch64 11.8` cannot be installed with `sysroot_linux-aarch64 2.28`
On several feedstocks I've seen CUDA on aarch rely on newer glibc symbols than 2.17.

For example, on the pycuda feedstock, this looks like
```
    from pycuda._driver import *  # noqa
ImportError: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by $PREFIX/lib/python3.10/site-packages/pycuda/../../../libcurand.so.10)
```
though interestingly it only happens for CUDA 11.8 and not 12.0 (on the faiss feedstock it was the other way around; in any case, it needs quite some [gymnastics](https://github.com/conda-forge/faiss-split-feedstock/blob/main/recipe/conda_build_config.yaml) to override).

However, when doing that for 11.8 (again example is [pycuda](https://github.com/conda-forge/pycuda-feedstock/pull/44)), I get:
```
Could not solve for environment specs
The following packages are incompatible
├─ nvcc_linux-aarch64 11.8.*  is installable and it requires
│  └─ sysroot_linux-aarch64 2.17.* , which can be installed;
└─ sysroot_linux-aarch64 2.28.*  is not installable because it conflicts with any installable versions previously reported.
```
So CUDA 11.8 may require symbols from glibc 2.27, but `nvcc` cannot be installed with a newer sysroot? It sounds to me like we should either loosen the version-constraint, or build two variants (one with 2.17, one with 2.28).

h-vetinari:
Closes #100 
Closes #107

Xref https://github.com/conda-forge/ctng-compilers-feedstock/pull/148


h-vetinari:
@conda-forge/nvcc, this should be ready, PTAL! :)

PS. If someone really wants to keep versions between 11.2 & 11.8, I don't mind, but AFAIU all those images are unsupported by now. If this gets merged I'm planning to remove the bits in conda-forge-ci-setup for <11.8 so we simply error on older versions.

kkraus14 (approved):

h-vetinari:
@conda-forge/nvcc, can we merge here? :)

PR #115, opened 2025-01-22
Use `pin_compatible` for `cudatoolkit` & `sysroot`

jakirkham:
Fix a package constraint bug where `sysroot` lacked an upper bound. Thus it is constrained within `>=2.17,<3.0a0`

Do this by adding pinned dependencies in `requirements/host` and then using `pin_compatible` to constrain these in `run_exports` and `run` (instead of rolling our own solution as before). This lines up with the logic `cuda-nvcc` currently uses

<hr>

Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -18,8 +18,9 @@ build:
 
 outputs:
   - name: nvcc_{{ cross_target_platform }}
-    script: linux/install_nvcc.sh     # [linux]
-    script: windows/install_nvcc.bat  # [win]
+    {% set script = "linux/install_nvcc.sh" %}     # [linux]
+    {% set script = "windows/install_nvcc.bat" %}  # [win]
+    script: {{ script }}
     build:
       ignore_run_exports_from:
         - {{ c_compiler }}_{{ cross_target_platform }}
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -7,7 +7,7 @@ package:
   version: {{ cuda_compiler_version }}
 
 build:
-  number: 25
+  number: 26
   skip: {{ not (compiler("cuda").startswith("nvcc") and cuda_compiler_version == "11.8") }}
 
 outputs:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,9 +1,3 @@
-{% if cuda_compiler_version in (None, "None", True, False) %}
-{% set cuda_major = 0 %}
-{% else %}
-{% set cuda_major = environ.get("cuda_compiler_version", "11.8").split(".")[0]|int %}
-{% endif %}
-
 {% if cross_target_platform is undefined %}
 {% set cross_target_platform = target_platform %}
 {% endif %}
@@ -17,7 +11,7 @@ build:
   skip: {{ not (compiler("cuda").startswith("nvcc") and cuda_compiler_version == "11.8") }}
 
 outputs:
-  - name: nvcc_{{ cross_target_platform }}
+  - name: {{ cuda_compiler }}_{{ cross_target_platform }}
     {% set script = "linux/install_nvcc.sh" %}     # [linux]
     {% set script = "windows/install_nvcc.bat" %}  # [win]
     script: {{ script }}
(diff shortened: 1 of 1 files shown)


jakirkham:

jakirkham:
@conda-forge-admin , please re-render

recipe/meta.yaml
```diff
@@ -13,33 +7,35 @@ package:
...
   skip: {{ not (compiler("cuda").startswith("nvcc") and cuda_compiler_version == "11.8") }}
 
 outputs:
-  - name: nvcc_{{ cross_target_platform }}
-    script: linux/install_nvcc.sh     # [linux]
-    script: windows/install_nvcc.bat  # [win]
+  - name: {{ cuda_compiler }}_{{ cross_target_platform }}
+    {% set script = "linux/install_nvcc.sh" %}     # [linux]
+    {% set script = "windows/install_nvcc.bat" %}  # [win]
+    script: {{ script }}
```
jakirkham, line 17: This fixes a linter hint with souschef above: https://github.com/conda-forge/nvcc-feedstock/pull/115#issuecomment-2608390342

recipe/meta.yaml
```diff
@@ -13,33 +7,35 @@ package:
...
 build:
-  number: 25
+  number: 26
   skip: {{ not (compiler("cuda").startswith("nvcc") and cuda_compiler_version == "11.8") }}
 
 outputs:
-  - name: nvcc_{{ cross_target_platform }}
-    script: linux/install_nvcc.sh     # [linux]
-    script: windows/install_nvcc.bat  # [win]
+  - name: {{ cuda_compiler }}_{{ cross_target_platform }}
```
jakirkham, line 14: This fixes a re-render issue where re-rendering was not adding `cuda-compiler` and in some cases disabling CI

Note the package name is exactly the same as before

recipe/meta.yaml
```diff
@@ -13,33 +7,35 @@ package:
...
       build:      # [win]
         - m2-sed  # [win]
       host:
-        - {{ stdlib("c") }}
         # Needed to symlink libcuda into sysroot libs.
         - {{ c_compiler }}_{{ cross_target_platform }} {{ c_compiler_version }}.*       # [linux]
         - {{ c_compiler }}_{{ cross_target_platform }}                                  # [win]
-        - sysroot_{{ cross_target_platform }} >={{ c_stdlib_version }}  # [linux]
+        - {{ c_stdlib }}_{{ cross_target_platform }} {{ c_stdlib_version }}             # [linux]
+        - {{ c_stdlib }}_{{ cross_target_platform }}                                    # [win]
```
jakirkham, line 33: This ensures we get `{{ stdlib("c") }}` in Linux and Windows cases with Linux being pinned to the version we intend to use as the minimum

recipe/meta.yaml
```diff
@@ -13,33 +7,35 @@ package:
...
+        - cudatoolkit {{ cuda_compiler_version }}
```
jakirkham, line 34: This adds `cudatoolkit` to `requirements/host` with a pin matching the version of this package. We can later use this to set the pinning on `cudatoolkit`

recipe/meta.yaml
```diff
@@ -13,33 +7,35 @@ package:
...
       run:
-        - cudatoolkit >={{ cuda_compiler_version }},<{{ cuda_major + 1 }}
-        - sed                                                           # [linux]
-        - sysroot_{{ cross_target_platform }} >={{ c_stdlib_version }}  # [linux]
+        - {{ pin_compatible("cudatoolkit") }}
```
jakirkham, line 36: Use `pin_compatible` with `cudatoolkit` to ensure it is pinned with the minimum from `requirements/host`. Also the default behavior is to allow any newer version as long as the major version is the same. So this is limited to CUDA 11 (as was the case before)

recipe/meta.yaml
```diff
@@ -13,33 +7,35 @@ package:
...
+        - sed                                                                           # [linux]
+        - {{ pin_compatible(c_stdlib ~ "_" ~ cross_target_platform) }}                  # [linux]
```
jakirkham, line 38: Also use `pin_compatible` with `{{ stdlib("c") }}` on Linux to ensure the minimum is pulled from `requirements/host` and only the same major version is allowed. The latter constraint was missed before

recipe/meta.yaml
```diff
@@ -13,33 +7,35 @@ package:
...
         strong:
-          - cudatoolkit >={{ cuda_compiler_version }},<{{ cuda_major + 1 }}
-          - sysroot_{{ cross_target_platform }} >={{ c_stdlib_version }}    # [linux]
+          - {{ pin_compatible("cudatoolkit") }}
+          - {{ pin_compatible(c_stdlib ~ "_" ~ cross_target_platform) }}  # [linux]
     requirements:
       build:      # [win]
         - m2-sed  # [win]
       host:
-        - {{ stdlib("c") }}
```
jakirkham, line 34: This is no longer needed as it is handled by the lines below with `{{ c_stdlib }}_{{ cross_target_platform }}`

recipe/meta.yaml
```diff
@@ -1,9 +1,3 @@
-{% if cuda_compiler_version in (None, "None", True, False) %}
-{% set cuda_major = 0 %}
-{% else %}
-{% set cuda_major = environ.get("cuda_compiler_version", "11.8").split(".")[0]|int %}
-{% endif %}
```
jakirkham, line 5: This is no longer needed as we now use `pin_compatible` to constrain `cudatoolkit` instead of rolling our own solution

recipe/meta.yaml
```diff
@@ -13,33 +7,35 @@ package:
...
+    script: {{ script }}
     build:
       ignore_run_exports_from:
         - {{ c_compiler }}_{{ cross_target_platform }}
       run_exports:
         strong:
-          - cudatoolkit >={{ cuda_compiler_version }},<{{ cuda_major + 1 }}
-          - sysroot_{{ cross_target_platform }} >={{ c_stdlib_version }}    # [linux]
+          - {{ pin_compatible("cudatoolkit") }}
+          - {{ pin_compatible(c_stdlib ~ "_" ~ cross_target_platform) }}  # [linux]
```
jakirkham, line 24: Also use `pin_compatible` in `run_exports` to add these constraints

jakirkham (commented):
This borrows the `{{ stdlib("c") }}` logic [from `cuda-nvcc`]( https://github.com/conda-forge/cuda-nvcc-feedstock/blob/4731c843f5132f327282c43498bd5cf500ed1315/recipe/meta.yaml#L55-L64 ), which is simpler and has worked reliably

Highlighted relevant changes below

